PR #17, opened 2022-06-01
Major refactor of code, Added layout features for overview page

SanjeevLakhwani:
## Dev Changes
- Upgraded webpack to v5 from v3
- Added webpack watch command to be able for a better developing experience
    - To use it, one can after running `npm run gobuild`, can run `npm run watch` on a different terminal from the same root folder
- Webpack now also compiles/bundles with original source code for better browser debugging experience
- Converted all the existing class based components to function based components
- Refactored all the code to be more optimised, logical and non-monolitihic
- Added redux toolkit instead of using plain redux and thunk, to have a more logical store.
    - Now a store slice is made with `createSlice` from redux toolkit, thunks are made with `createAsyncThunk` and the depreciated `createStore` command has been switched to the recommended `configureStore` command
-  Constants and Utils were also refactored for appropriate use
### Overview Page
- Charts are now displayed in a card component
    -  Charts now do not have a title property but the cards have the title property now
- Cards in their header have up and down arrow to rearrange them by 1
- Cards in their header have a + button for to bring manage charts drawer
- Cards in their header have a x button to remove it
- The arrangement of charts, and whether they are displayed is persisted and stored in the localstorage
- The last card is the + card which also brings up the Manage Charts Drawer
- The manage charts drawer has the title of all the charts as a list
    - The list items are draggable and their rearrangement is synchronous to the arrangement of charts on the overview page
    - The list items also have a checkbox to manage whether the chart is displayed or not
- The overview page has an individuals count in header
### Search Page
The search page on the outside doesn't have many changes but massively refactored internally
- After reaching the query parameter selection limit, the other checkboxes are now disabled
- Units how now been moved to the description tooltip
- The "Get Results" button has been changed to have an internal loader and different UI


src/js/utils/DataUtils.js
```diff
@@ -0,0 +1,138 @@
...
+      if (intKey < leftTaper) {
+        const tlkey = `< ${leftTaper}`;
+        if (taperLeft == undefined) {
+          taperLeft = { x: tlkey, y: inputData[key] };
+        } else {
+          taperLeft.y += inputData[key];
+        }
+      } else if (intKey >= rightTaper + binSize) {
+        const trkey = `>= ${rightTaper + binSize}`;
+        if (taperRight == undefined) {
```
ppillot, line 64: same as above

src/js/utils/DataUtils.js
```diff
@@ -0,0 +1,138 @@
...
+const parseDataGeneral = (inputData) =>
+  Object.keys(inputData).map((key) => ({ x: key, y: inputData[key] }));
+
+export const binMissingData = (data) => {
+  const missingObject = data.filter((obj) => obj.x === 'missing');
+
+  if (missingObject.length > 0) {
+    // fancy one liner to splice the object out and re-append it
+    data.push(data.splice(data.indexOf(missingObject[0]), 1)[0]);
+  }
```
ppillot, line 110: nit. can be simplified by calling `data.indexOf` in a first step and avoiding the filter part. Should be more performant also as `.filter` parses the whole array, when `indexOf` stops as soon as the "missing" bin is found

src/js/utils/localStorage.js
```diff
@@ -0,0 +1,41 @@
+export const verifyData = (nArr, oArr) => {
+  if (nArr.length !== oArr.length) {
+    return false;
+  }
+
+  return oArr.every(
+    (e) =>
+      nArr.findIndex(
```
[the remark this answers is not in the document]
ppillot: Also the construct .findIndex() !== -1 could be made more readable by replacing it with `.includes()`

src/js/utils/localStorage.js
```diff
@@ -0,0 +1,41 @@
...
+  }
+};
+
+export const getValue = (key, defaultVal, verifyFunc) => {
+  try {
+    const serializedState = localStorage.getItem(key);
+    if (serializedState === null) {
+      return defaultVal;
+    }
+    const unsearlizedState = JSON.parse(serializedState);
```
ppillot, line 29: spelling (`unsearlized...`)

src/js/utils/localStorage.js
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    return unsearlizedState;
+  } catch (err) {
+    return defaultVal;
+  }
+};
+
+export const convertSequenceAndDisplayData = (arr) =>
+  arr.map((e) => ({ name: e.name, isDisplayed: e.isDisplayed }));
```
ppillot, line 41: Sugg: you might enforce the cast to boolean with `isDisplayed: !!e.isDisplayed`

src/js/utils/queryUtils.js
```diff
@@ -0,0 +1,32 @@
...
+  ...otherProps
+}) => {
+  if (type === 'string') {
+    if (format === 'date') return queryTypes.DATE;
+    else if (otherProps.enum) return queryTypes.SELECT;
+    else return queryTypes.INPUT;
+  } else if (type === 'number') {
+    if (is_range) return queryTypes.RANGE;
+    else return queryTypes.NUMBER_INPUT;
+  } else return queryTypes.INPUT;
```
ppillot, line 17: FOr legibility, indent this statement or wrap it in curly braces to put it at the same level as its counterparts

src/js/features/data.js
```diff
@@ -0,0 +1,133 @@
...
+      const localValue = getValue(LS_CHARTS_KEY, convertedData, (val) =>
+        verifyData(val, convertedData)
+      );
+
+      const temp = [...state.chartData];
+
+      state.chartData = localValue.map((e) => ({
+        ...temp.find((v) => v.name === e.name),
+        isDisplayed: e.isDisplayed,
+      }));
```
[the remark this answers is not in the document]
SanjeevLakhwani: The number of charts is going to be small anyways so nothing to worry about

src/js/features/data.js
```diff
@@ -0,0 +1,133 @@
...
+
+      convertedData = convertSequenceAndDisplayData(state.chartData);
+
+      saveValue(LS_CHARTS_KEY, convertedData);
```
gsfk, line 123: I'm not an expert on redux toolkit but this looks an awful lot like you have side effects in your reducers. These should be in action creators instead.

src/js/utils/DataUtils.js
SanjeevLakhwani: on analyzing, the function is redundant as it is always binned in the data already

SanjeevLakhwani: I am just gonna remove the function

SanjeevLakhwani: on analyzing the function, the only thing it does is to move it to the end of the data, so im just gonna create a function to do that

gsfk (changes requested):
This is great! It's _much_ more attractive than it was. Thank you.

There were a few code puzzles for me, but I tried to focus on tests, since Paul has been through the code so thoroughly already. 

A general issue that came up in tests is that there's a brittle dependency on the katsu version: it's not resilient to minor additions to katsu's config file, and tends to crash on any unexpected changes to the overview response. This is also a katsu issue, the config file itself could probably be more robust. 


A few design comments:

- The giant "plus sign" card seems awfully huge for a config menu, it's also misleading since it does more than add charts, it also removes them, and does whatever else might go on in the drawer. (The drawer is great!)

- Why is there a "manage charts" button on every chart?

- Do users really need to re-order charts? And if yes, is the best solution buttons on every chart? Can this be done in the drawer? Or by making them draggable? I don't know what the most intuitive solution is. 

- Tooltips for bar chart bins now only show the labels, ideally they should show the _range_ covered by the bar rather than just the label. 

- Pie charts are cropped a bit, eg:

Screenshot from 2022-06-03 09-49-54

src/js/utils/localStorage.js
SanjeevLakhwani: already taken care of with checking the type of isDisplayed to be boolean

src/js/components/TabbedDashboard.js
```diff
@@ -1,38 +1,59 @@
...
+  const tabBarStyle = { marginBottom: '20px' };
 
-  const overviewTabTitle = <p style={tabTitleStyle}>Overview</p>
-  const searchTabTitle = <p style={tabTitleStyle}>Search</p>
+  const overviewTabTitle = <p style={tabTitleStyle}>Overview</p>;
+  const searchTabTitle = <p style={tabTitleStyle}>Search</p>;
+
+  const { overview, queryParameterStack, isFetchingData } = useSelector(
+    (state) => state.data
+  );
```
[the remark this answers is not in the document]
SanjeevLakhwani: Oops well it is great to know that that is the case but umm, overview and queryParameterStack were not even required in this case, that was just old cade I forgot to remove. So now there is only one, isFetchingData.

src/js/components/overview/Drawer/ChartTree.js
```diff
@@ -0,0 +1,78 @@
...
+    if (dropPosition === -1) {
+      ar.splice(i, 0, dragObj);
+    } else {
+      ar.splice(i + 1, 0, dragObj);
+    }
+
+    dispatch(rearrange(data.map((e) => e.key)));
+  };
+
+  const checkedKeys = useSelector((state) =>
```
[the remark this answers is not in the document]
SanjeevLakhwani: yeah that is my fault, I copied the code from antd website and didnt care myself to refactor it. It collides with pauls comments too, therefore im gonna mark it as resolved

src/js/constants/overviewConstants.js
```diff
@@ -0,0 +1,49 @@
...
+//   "#fdae61",
+//   "#abd9e9",
+//   "#74add1",
+//   "#4575b4",
+// ]
+
+export const BAR_CHART_FILL = '#4575b4';
+export const BAR_CHART_MISSING_FILL = '#bbbbbb';
+
+export const LS_CHARTS_KEY = 'charts';
```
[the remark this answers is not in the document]
SanjeevLakhwani: Local Storage Charts Key, it is just the key we are using to store the chart sequence data in local storage

src/js/features/data.js
SanjeevLakhwani: I know but the otherway was to have an increased code complexity, which i wanted to kind of run away from. I would look into it.

src/js/components/overview/Drawer/ChartTree.js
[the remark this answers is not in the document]
SanjeevLakhwani: yup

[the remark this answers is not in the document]
SanjeevLakhwani: an array of keys (name) that are checked, this is how antd processes them

src/js/features/data.js
SanjeevLakhwani: going through the code again , I realized that I could have done that in the thunk itself

src/js/constants/overviewConstants.js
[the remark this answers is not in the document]
gsfk: consider renaming to LOCAL_STORAGE_CHART_KEY or something more clear

src/js/utils/DataUtils.js
```diff
@@ -19,52 +15,51 @@ const parseDateData = (inputData) => {
...
 };
 
 const parseNumericalData = (inputData, leftTaper, rightTaper, binSize) => {
   let formattedData = [];
   let taperLeft, taperRight;
 
-  Object.keys(inputData).forEach((key, index, arr) => {
+  Object.entries(inputData).forEach(([key, value], index, arr) => {
     const intKey = parseInt(key);
     if (intKey != NaN) {
```
ppillot, line 49: this is always false (even `NaN === NaN` is false)
use `!Number.isNaN(intKey)`

src/js/utils/DataUtils.js
```diff
@@ -118,21 +103,19 @@ export const parseData = ({ data, properties: props }) => {
...
-  formattedData = binMissingData(formattedData);
+  const missingIndex = formattedData.findIndex((e) => e.x === 'missing');
+  if (missingIndex !== -1)
+    formattedData.push(formattedData.splice(missingIndex, 1)[0]);
 
   // order categories according to config file, where applicable
-  if (props.enum) {
-    formattedData = orderCategories(formattedData, props.enum);
-  }
+  if (props.enum) formattedData = orderCategories(formattedData, props.enum);
```
ppillot, line 118: Maybe this should be moved before the "missing to the end" part? I am afraid we are not guaranteed tiah the "missing" category will be at the end??

src/js/utils/localStorage.js
```diff
@@ -26,12 +25,12 @@ export const getValue = (key, defaultVal, verifyFunc) => {
     if (serializedState === null) {
       return defaultVal;
     }
-    const unsearlizedState = JSON.parse(serializedState);
-    if (!verifyFunc(unsearlizedState)) {
+    const unsearializedState = JSON.parse(serializedState);
```
ppillot, line 28: spelling? `unserializedState`

ppillot (changes requested):
Thanks for these modifications. It's even more clear now.
Please fix the `== NaN`issue

PR #21, opened 2022-06-15
Fix handle missing extra properties

ppillot:
How to test:
Using the following config.json for Katsu (notice: it does not contain `extra_properties`), load the current main branch for bento_public. Check that you can reproduce the bug where no data is displayed.
Update your repo to this branch and check that everything is displayed correctly.

```json
{
  "age": {
    "type": "number",
    "title": "Age",
    "bin_size": 10,
    "is_range": true,
    "queryable": true,
    "taper_left": 40,
    "taper_right": 60,
    "units": "years",
    "minimum": 0,
    "description": "Age at arrival"
  },
  "sex": {
    "type": "string",
    "enum": [
      "Male",
      "Female"
    ],
    "title": "Sex",
    "queryable": true,
    "description": "Sex at birth"
  },
  "experiment_type": {
    "type": "string",
    "title": "Experiment Types",
    "queriable": false,
    "description": "Types of experiments performed on a sample"
  }
}
```

### DIFF
--- a/src/js/features/data/makeGetDataRequest.js
+++ b/src/js/features/data/makeGetDataRequest.js
@@ -16,6 +16,17 @@ import {
 import { LOCALSTORAGE_CHARTS_KEY } from '../../constants/overviewConstants';
 import { addQueryableFields } from '../query';
 
+/**
+ * Check wether a given property has the shape of a chart configuration, i.e.
+ * an object where every key is associated with a numerical value
+ * @param {object} prop
+ * @returns
+ */
+const isChartConfig = (prop) => {
+  if (typeof prop !== 'object') return false;
+  return Object.values(prop).every((v) => typeof v === 'number');
+}
+
 export const makeGetDataRequest = createAsyncThunk(
   'data/getConfigData',
   async (_ignore, thunkAPI) => {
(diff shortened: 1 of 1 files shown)


src/js/features/data/makeGetDataRequest.js
```diff
@@ -28,14 +39,14 @@ export const makeGetDataRequest = createAsyncThunk(
       // converting fields to usable form
       let fields = {
         ...queryParameterStack,
-        ...queryParameterStack.extra_properties,
+        ...(queryParameterStack.extra_properties ?? {}),
       };
-      delete fields.extra_properties;
+      if (fields.extra_properties) delete fields.extra_properties;
```
gsfk, line 44: This _works_, but is the goal here just to remove the extra nesting around what's inside extra_properties? This can be done with destructuring: 

```
const {extra_properties, ...everything_else} = queryParameterStack
let fields = {...everything_else, ...extra_properties}
```

which I find easier to understand, and requires no extra checks to see if an extra_properties field is present.

gsfk (approved):
Looks good. I offered up one suggestion for less complex code.

src/js/features/data/makeGetDataRequest.js
ppillot: This looks nice!
I've never realized that `undefined` could be spread. Thanks!

gsfk: haha! I didn't notice it either until today.

PR #27, opened 2022-08-02
This contains changes adapting bento public to changes in katsu's responses

SanjeevLakhwani:
Changes
- Overview and Search tabs now have sections
- Data representation format has changed throughout (internally) to accommodate with Katsu
- Overview and Search tabs are now more responsive to screen size changes
- Search results now have a new ui 
- Search result is invalidated if any change occurs to the search options


ppillot (approved):
I've done a cursory look through the code, mainly focusing on understanding the changes and looking for possible clarifications.
Thank you for having taken the time to rewrite a meaningful commit history and overall to the attention to the details that make code clearer (e.g. order of the imports, cleaning of unused variables, code compartmentalization between files, better formatting settings).
Branch has been rebased on main to avoid conflicts with now superseded fixes that had been made on the main branch.
The tests have shown two issues that I've addressed in two different commits:
- one is related with the communication with the server that seems to throttle in some circumstances (Usage of Django dev server?). Sequential calls are made instead of concurrent at initialization.
- one is related with the positioning of the individual counts which was fixed, leading to be scrolled over by the charts when the user scrolls.

PR #33, opened 2022-09-16
tooltip for charts, updates on provenance

SanjeevLakhwani:
- Added a tooltip on title of chart cards providing description
- Added primary publications to provenance
- Description fields with no data would not be displayed on provenance
- Removed clipping on description fields by separating lines


ppillot:
Has there been a change in the format of the DATS file? I have encountered some error messages complaining about a missing `roles` property?

noctillion (commented):
Everything looks good! .. starting the testing asap.

brouillette:
Current implementation crashes on the `Provenance` tab when looping over the `dats_file.creators` where a `creator` does not have the field name being requested --

image

-- in this case, the `roles`

image


Further observation of the code reveals that the possibility of mishandling the `abbreviation` field as well, depending on the data. See https://github.com/c3g/ichange_scripts/blob/main/ICHANGE_DATS.json and you'll notice not all creators have `abbreviation`.

Please consider adding null/undefined checks for both the column names (i.e. the json field name) and the values being looped over.

brouillette:
Works like a charm -- Thank you!

PR #34, opened 2022-10-07
Search chart resp

SanjeevLakhwani:
- Created a utility function out of serializing charts
- Added biosample and experiment counts and charts to the search result

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "mini-css-extract-plugin": "^2.4.5",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
+    "react-icons": "^4.4.0",
     "react-json-formatter": "^0.2.1",
     "react-notify-bootstrap": "^1.0.23",
     "react-redux": "^7.2.6",
--- a/src/js/components/Search/SearchResults.js
+++ b/src/js/components/Search/SearchResults.js
@@ -33,7 +33,7 @@ const SearchResults = () => {
             <Space direction="vertical" size="middle" style={{ display: 'flex' }}>
               <Statistic
                 title="Individual Count"
-                value={isValid && status === 'count' ? (status === 'count' ? count : message) : '----'}
+                value={isValid ? (status === 'count' ? count : message) : '----'}
                 valueStyle={{ color: '#1890ff' }}
                 prefix={<TeamOutlined />}
               />
--- a/src/js/components/Search/SearchResults.js
+++ b/src/js/components/Search/SearchResults.js
@@ -1,11 +1,17 @@
 import React from 'react';
 import { useSelector } from 'react-redux';
-import { Card, Statistic } from 'antd';
+import { Card, Col, Empty, Row, Statistic, Typography } from 'antd';
 import { TeamOutlined } from '@ant-design/icons';
+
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -26,6 +26,7 @@
     "mini-css-extract-plugin": "^2.4.5",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
+    "react-icons": "^4.4.0",
```
ppillot, line 29: How much does this add-up to the overall payload? Is it only the icons that are being requested or is the whole set downloaded to the client?

src/js/components/Search/SearchResults.js
```diff
@@ -15,13 +27,47 @@ const SearchResults = () => {
...
+        <Row gutter={16}>
+          <Col span={6}>
+            <Space direction="vertical" size="middle" style={{ display: 'flex' }}>
+              <Statistic
+                title="Individual Count"
+                value={isValid && status === 'count' ? (status === 'count' ? count : message) : '----'}
+                valueStyle={{ color: '#1890ff' }}
+                prefix={<TeamOutlined />}
+              />
+              <Statistic
```
ppillot, line 40: I think we considered this component and the following one to be displayed only when there is enough data?

src/js/utils/chart.js
```diff
@@ -0,0 +1,3 @@
+export const serializeChartData = (chartData) => {
+  return chartData.map(({ label, value }) => ({ x: label, y: value }));
```
ppillot, line 2: Nice refactor!

src/js/components/Search/SearchResults.js
```diff
@@ -15,13 +27,47 @@ const SearchResults = () => {
...
   return (
     <div style={wrapperStyle}>
-      <Card style={{ borderRadius: '10px' }}>
-        <Statistic
-          title="Count"
-          value={isValid ? (status === 'count' ? count : message) : '----'}
-          valueStyle={{ color: '#1890ff' }}
-          prefix={<TeamOutlined />}
-        />
+      <Card style={{ borderRadius: '10px', padding: '10px 33px', width: '900px' }}>
```
ppillot, line 30: Minor: the 900px width can be limiting on a phone screen... Maybe `max-width` would be more appropriate to address the issue of wide computer screens, and otherwise on narrower screen it should allow the content (i.e. the charts) to wrap.
But maybe this is limited by the `Col` component anyway...

ppillot (approved):
Some minor suggestions that can be investigated or not.
Good refactor of the code for reusability.
Can be merged if functional tests are successful.

package.json
SanjeevLakhwani: it only adds the icons that requested to the payload

src/js/components/Search/SearchResults.js
SanjeevLakhwani: the value wouldn't be returned by katsu, so it cannot really show anything

src/js/utils/chart.js
SanjeevLakhwani: ty

brouillette:
Tested and approved!

PR #35, opened 2022-10-16
Internationalization

SanjeevLakhwani:
Internatlonalized bento public, it now has translation for french and english. click on the button on site header to switch languages


gsfk (commented):
Nice! A few questions: 
- the button to change language shows the _current_ language (ie: on the English page it shows `en`). Isn't the usual practice the other way around? (ie `fr` on the English page, so that the button is more clearly a "change language" button)
- shouldn't changing the language affect the page's laguage attribute? It's always `en` no matter what language is chosen.

brouillette:
LGTM!

brouillette (approved):
👍

PR #36, opened 2022-10-31
Styling changes

SanjeevLakhwani:
- moved the search results to the top
- moved get data to the side as a floating button
- Bento public logo and the text next to it now do not overlap on zoom or small screen
- made overview charts response and responds well to screen size change
- made default chart load for a section to 3

### DIFF
--- a/src/js/components/Overview/OverviewDisplayData.js
+++ b/src/js/components/Overview/OverviewDisplayData.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import { useDispatch } from 'react-redux';
-import { Row, Col } from 'antd';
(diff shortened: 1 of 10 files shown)


gsfk (approved):
Look good, with two small comments:

- On very wide screens, charts don't centre properly if there are only two charts in a category (it hugs to the left, leaving room for a phantom third chart). This happens even if none of the categories has more than two charts.

- Search results are at the top, but the button to launch a search is all the way at the bottom of the page. The main annoyance here is that the list of search fields is too big, but that's outside the scope of this pr.

dbujold:
Ready to merge!

PR #38, opened 2022-11-11
Added footer, styling & translation changes

SanjeevLakhwani:
- Added a sitewide footer
- Added rounded corners to provenance card
- Added translation to charts drawer
- Added translation to chart description tooltip
- Separated default and non-default translations


gsfk (approved):
Looks good. I find the version of the footer a bit uglier than the original bento_web version.

PR #42, opened 2022-11-16
v0.7.0 rc-1

brouillette:
- fixed translation terms
- client now relies on /config http call to obtain client name
- no more need to have 2 .env configuration file
- ci docker image build


gsfk (commented):
Tested locally and seems good, although I didn't look at the CI stuff.  In particular "client name" works well.

I wasn't sure about `build-dev`, since `watch` didn't seem to work, it was still necessary to run watch in a second terminal.

brouillette:
Thanks
The `build-dev` command builds the JS without obfuscating the code so the files can be debugged in the browser

PR #43, opened 2022-11-18
Fix translation issues, tweak styling, and upgrade to Ant Design 5

SanjeevLakhwani:
- fixed translation issue on tooltip
- fixed translation issue on insufficient data message
- header buttons rounded
- Upgraded to antdv5
- optimized tabpane rendering to be more scalable

### DIFF
--- a/src/js/components/Overview/charts/BentoPie.js
+++ b/src/js/components/Overview/charts/BentoPie.js
@@ -41,8 +41,9 @@ function BentoPie({ data, height }) {
 
   // Changing immutable data to mutable data
   data = [...data];
-  // remove "missing" field if zero
+
   data = data.filter((e) => !(e.x === 'missing'));
+  data = data.map((d) => ({ ...d, x: t(d.x) }));
 
   // combining sections with less than OTHER_THRESHOLD
   const sum = data.reduce((acc, e) => acc + e.y, 0);
@@ -120,7 +121,7 @@ function BentoPie({ data, height }) {
         <path d={`M${sx},${sy}L${mx},${my}L${ex},${ey}`} stroke={fill} fill="none" />
         <circle cx={ex} cy={ey} r={2} fill={fill} stroke="none" />
         <text x={ex + (cos >= 0 ? 1 : -1) * 12} y={ey + 3} textAnchor={textAnchor} style={currentTextStyle}>
-          {labelShortName(t(name))}
+          {labelShortName(name)}
         </text>
         <text x={ex + (cos >= 0 ? 1 : -1) * 12} y={ey} dy={14} textAnchor={textAnchor} style={countTextStyle}>
           {`(${t(payload.value)})`}
--- a/src/js/components/Overview/charts/BentoPie.js
+++ b/src/js/components/Overview/charts/BentoPie.js
@@ -39,11 +39,9 @@ function BentoPie({ data, height }) {
 
   const [activeIndex, setActiveIndex] = useState(undefined);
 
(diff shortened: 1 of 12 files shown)


src/js/components/Overview/charts/BentoPie.js
```diff
@@ -41,8 +41,9 @@ function BentoPie({ data, height }) {
 
   // Changing immutable data to mutable data
   data = [...data];
-  // remove "missing" field if zero
+
   data = data.filter((e) => !(e.x === 'missing'));
+  data = data.map((d) => ({ ...d, x: t(d.x) }));
```
davidlougheed, line 46: I would avoid these reassignments; imo its not any less readable in the form of:

```js
data = [...data]
    .filter((e) => !(e.x === 'missing'))
    .map((d) => ({ ...d, x: t(d.x) }));
```
with indenting. In general avoiding mutation of the variable (reference) value is good, although this does get mutated later a bunch anyway...

src/js/components/Overview/charts/BentoPie.js
```diff
@@ -120,7 +121,7 @@ function BentoPie({ data, height }) {
         <path d={`M${sx},${sy}L${mx},${my}L${ex},${ey}`} stroke={fill} fill="none" />
         <circle cx={ex} cy={ey} r={2} fill={fill} stroke="none" />
         <text x={ex + (cos >= 0 ? 1 : -1) * 12} y={ey + 3} textAnchor={textAnchor} style={currentTextStyle}>
-          {labelShortName(t(name))}
+          {labelShortName(name)}
```
davidlougheed, line 124: for my own understanding, why is this now untranslated?

src/js/components/TabbedDashboard.js
```diff
@@ -30,27 +29,48 @@ const TabbedDashboard = () => {
...
+    },
+    {
+      title: 'Provenance',
+      content: <ProvenanceTab />,
+      loading: false,
+      key: 'provenance',
+    },
+  ];
+
+  const MappedTabPanes = TabPanes.map(({ title, content, loading, key }) => ({
```
davidlougheed, line 66: this shouldn't be name-formatted like it's a component

src/js/components/TabbedDashboard.js
```diff
@@ -30,27 +29,48 @@ const TabbedDashboard = () => {
...
   const isFetchingSearchFields = useSelector((state) => state.query.isFetchingFields);
+  const individuals = useSelector((state) => state.data.individuals);
+
+  const individualCount = (
+    <Title level={5}>
+      {t('Individuals')}: {individuals}
+    </Title>
+  );
+
+  const TabPanes = [
```
davidlougheed, line 45: this shouldn't be name-formatted like it's a component

davidlougheed (changes requested):

src/js/components/Overview/charts/BentoPie.js
SanjeevLakhwani: It is translated earlier so no need after

davidlougheed: maybe put a comment mentioning where this comes from/where it gets translated to make it easy to understand in the future then?

SanjeevLakhwani:
Dealt with the requested changes!

davidlougheed (approved):

davidlougheed:
this should, when released, be versioned 0.7.0 when combined with Brennan's PR

PR #46, opened 2022-12-12
Changes 4

SanjeevLakhwani:
- converted overlay button to floatbutton element with tooltip
- Added terms of use in footer
- Added about section on overview
- Translation are now fetched from static sources made by webpack


brouillette:
LGTM

PR #52, opened 2022-12-22
Chart improvements

davidlougheed:
* Show more labels for bar charts with few categories
* Increase X axis label distance to avoid some overlaps


gsfk (approved):
Looks better! There's now more overlapping tick labels and x-axis labels, but this is a separate issue we can handle elsewhere.

PR #55, opened 2023-01-06
Dockerfile updates, configurable INTERNAL_PORT, /service-info endpoint

davidlougheed:


src/main.go
```diff
@@ -34,19 +48,41 @@ type QueryParameter struct {
...
+	}
+
+	var packageJsonContents map[string]interface{}
+	err = json.Unmarshal(packageJson, &packageJsonContents)
+	if err != nil {
+		fmt.Println("Error parsing package.json")
+		os.Exit(1)
+	}
+
+	var version = fmt.Sprintf("%s", packageJsonContents["version"])
```
v-rocheleau, line 73: Disclaimer: I'm a golang noob. Why use fmt.Sprintf here if the returned string is equal to `packageJsonContents["version"]` ?

davidlougheed: I thought I had a reason at some point, but I think it must have been an issue elsewhere triggering an IDE warning - anyway I'll remove it since you're right, it's redundant

brouillette (approved):
LGTM!
image

```
/service-info

{
"bento": {
"serviceKind": "bento"
},
"contactUrl": "mailto:info@c3g.ca",
"id": "",
"name": "Bento Public",
"organization": {
"name": "C3G",
"url": "https://www.computationalgenomics.ca/"
},
"type": {
"artifact": "public",
"group": "ca.c3g.bento",
"version": "0.9.0"
},
"version": "0.9.0"
}
```

brouillette (approved):

PR #56, opened 2023-01-07
[WIP] feature 1264

brouillette:
Search: When changing parameters, refresh the Count automatically instead of requiring the user to click "Get Data"


davidlougheed (changes requested):
in general i've suggested a couple directions to avoid prop-drilling with the dispatch function

brouillette (commented):
thanks
much cleaner now

src/js/components/Search/SelectOption.js
```diff
@@ -1,20 +1,33 @@
...
+const SelectOption = ({ id, isChecked, options, optionalDispatch}) => {
   const { t } = useTranslation();
   const dispatch = useDispatch();
 
   const [value, setValue] = useState(options[0]);
 
+  const isFirstRun = useRef(true);
   useEffect(() => {
     if (isChecked) {
       dispatch(addQueryParam({ id, value }));
```
v-rocheleau, line 17: If the user un-checks this option, won't the query param be out of sync?

src/js/components/Search/SelectOption.js
```diff
@@ -1,20 +1,33 @@
...
     }
+
+    // to prevent everything between the lines '--'
+    // at component initialization
+    if (isFirstRun.current) {
+      isFirstRun.current = false;
+      return;
+    }
+    // --
+    if (optionalDispatch != undefined) {
```
v-rocheleau, line 27: We could use the truthy form `if (optionalDispatch) {...}` to handle unexpected types

davidlougheed: also fwiw `!=` should never be used in JS anyway, always `!==`

src/js/components/Search/SelectOption.js
brouillette: Nope
This 'first-run' is only triggered upon first page and component load

v-rocheleau (approved):

davidlougheed (approved):

PR #57, opened 2023-01-24
v0.9.1 rc-1

brouillette:
Search page patches
- added 'back to top' button
- added 'loading' status to results card
- patched card height jumpiness

### DIFF
--- a/src/js/components/Search/Search.js
+++ b/src/js/components/Search/Search.js
@@ -1,5 +1,6 @@
 import React, {useEffect} from 'react';
-import { Row, Typography, Space } from 'antd';
+import { Row, Button, Typography, Space } from 'antd';
+import { UpOutlined } from '@ant-design/icons';
 import { useSelector,useDispatch  } from 'react-redux';
 import { useTranslation } from 'react-i18next';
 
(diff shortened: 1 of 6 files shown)


src/js/components/Search/Search.js
```diff
@@ -33,6 +39,17 @@ const Search = () => {
...
       </Row>
+      <Button
+        type="primary"
+        onClick={scrollToTop}
+        size="large"
+        shape="round"
+        loading={isFetchingData}
+        className="floating-search"
+      >
+        <UpOutlined />
```
davidlougheed, line 50: use `<Button icon={<UpOutlined />} ...>...` instead

src/js/components/Search/Search.js
```diff
@@ -13,11 +14,16 @@ const Search = () => {
   const { t } = useTranslation();
   const dispatch = useDispatch();
   const searchSections = useSelector((state) => state.query.querySections);
+  const isFetchingData = useSelector((state) => state.query.isFetchingData);
 
   useEffect(() => {
     dispatch(makeGetKatsuPublic());
   }, []);
 
+  const scrollToTop = () => {
```
davidlougheed, line 23: move this outside of component (so below the last import line) to avoid redefining it every render

davidlougheed (changes requested):
minor changes requested; other than those looks good!

gsfk (approved):
Works well for me. Had to make some Dockerfile changes to get it run on 2.8, but that's outside the scope of this pr.

v-rocheleau (approved):
Looks good.
UX comment: maybe only display "back to top" when scroll position is not already at the top

davidlougheed (approved):

PR #62, opened 2023-02-22
Refactoring; development image + security

davidlougheed:
* chore: make development image
* chore: improve image security (don't run as root)
* chore: update dependencies
* refact: switch to Go module format
* refact: reduce duplicate code for Katsu calls
* fix: deprecated io calls


run.dev.bash
v-rocheleau, line 1: In BentoV2 I can't get the bento_public to work properly in "local" mode, only get 404 responses when trying to load the public url.

v-rocheleau: Logs from the local container:

> {"time":"2023-03-20T14:59:59.748131628Z","id":"","remote_ip":"172.18.0.2","host":"bentov2.local","method":"GET","uri":"/","user_agent":"Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/109.0.0.0 Safari/537.36","status":404,"error":"code=404, message=Not Found","latency":25698,"latency_human":"25.698µs","bytes_in":0,"bytes_out":24}

v-rocheleau: Also this setup is looking pretty funky to me haha, didn't know nodemon could recompile go.
I think that caddy has some flags for stuff like this, if we want to use that instead of go one day.

package.json
```diff
@@ -14,54 +14,46 @@
...
-    "@types/react": "^18.0.28",
-    "@types/react-dom": "^18.0.10",
-    "@types/recharts": "^1.8.24",
-    "antd": "^5.0.0",
-    "axios": "^0.24.0",
+    "@ant-design/icons": "^5.0.1",
+    "@reduxjs/toolkit": "^1.9.3",
+    "antd": "^5.2.2",
+    "axios": "^1.3.4",
     "bento-charts": "^1.0.7",
```
v-rocheleau, line 21: Bento-charts changes from [this](https://github.com/bento-platform/bento_public/pull/64) pr need to be double checked. Error occuring in bento_public when displaying data charts in overview. Very likely due to React dependency errors

Problem:
- The bento-charts [repo](https://github.com/bento-platform/Bento-Charts) has to package.json
- Oddly, v1.0.7 was published with a package.json that includes conflicting dependencies with this branch of bento_public (react especially)
- Bento-charts should specify [peerDependencies](https://nodejs.org/ko/blog/npm/peer-dependencies) instead of dependencies for compatibility

v-rocheleau: Ongoing PR [here](https://github.com/bento-platform/Bento-Charts/pull/1) .
Waiting for review and NPM_TOKEN to be added to the Actions secrets for CI publish.

v-rocheleau: Testing with new package 1.0.8 before creating 2.0.0

v-rocheleau: Updated to 2.0.0 release.

Dockerfile
```diff
@@ -10,39 +10,34 @@ RUN mkdir -p build/www
 RUN npm run build-dev
```
v-rocheleau, line 10: Not `npm run build`?

run.dev.bash
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+export BENTO_PUBLIC_STATIC_FILES_PATH=/bento-public/build/www
+
+cd /bento-public || exit
+
+ls -la
+
```
v-rocheleau, line 8: Think we should add `npm run build-dev` here, fixed my issue where I only got 404 if I ran in local.
main.go uses a static file middleware to serve files in `build/www`, but this directory is ignored by git.
Adding `npm run build-dev` created the directory and files, after that no issues.

v-rocheleau (changes requested):
Docker comments, otherwise looks good!

Dockerfile
davidlougheed: true

v-rocheleau (approved):
Looks good!

PR #66, opened 2023-03-20
Added types to the subsection

SanjeevLakhwani:


package.json
```diff
@@ -8,7 +8,7 @@
     "build-dev": "webpack --mode=development",
     "watch": "webpack --mode=development --watch",
     "gobuild": "./buildandrun.sh",
-    "lint": "npx eslint src/js/**"
+    "lint": "npx eslint \"src/js/**/*.{js,jsx}\""
```
v-rocheleau, line 11: We should also lint the .ts files

src/js/features/data/makeGetDataRequest.thunk.ts
```diff
@@ -0,0 +1,53 @@
...
+import { verifyData, saveValue, getValue, convertSequenceAndDisplayData } from '@/utils/localStorage';
+
+import { LOCALSTORAGE_CHARTS_KEY } from '@/constants/overviewConstants';
+import { serializeChartData } from '@/utils/chart';
+import { LocalStorageData, Sections } from '@/types/Data.interface';
+import { Chart, Counts, OverviewResponse } from '@/types/OverviewResponse.interface';
+import { printAPIError } from '@/utils/error.util';
+
+export const makeGetDataRequestThunk = createAsyncThunk<
+  { sectionData: Sections; counts: Counts },
```
v-rocheleau, line 14: We could declare these anonymous types to improve readability

src/js/types/Data.interface.ts
v-rocheleau, line 1: We can assume all files under src/js/types are for types and interfaces, so I dont think it is informative to have this file naming pattern here

src/js/types/OverviewResponse.interface.ts
v-rocheleau, line 1: Same, file name

src/js/types/util.types.ts
v-rocheleau, line 1: Same, file name

v-rocheleau (changes requested):

package.json
SanjeevLakhwani: Yes but I want to wait till I convert everything to typescript just cause, there are like a couple situations where there are unused exports/variables that would be picked up by the linter but we need the full typescript conversion for them to go away. Otherwise it is not that bad cause my IDE also tries to lint stuff and I have prettier.

PR #68, opened 2023-03-31
fix: webpack in dev container

v-rocheleau:
Fix dev container to reload webpack changes

### DIFF
--- a/run.dev.bash
+++ b/run.dev.bash
@@ -13,9 +13,10 @@ npm install
 npm run build-dev
 
 # Run nodemon as a watcher to recompile Go
-npx nodemon main.go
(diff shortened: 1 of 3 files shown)


run.dev.bash
```diff
@@ -13,9 +13,10 @@ npm install
 npm run build-dev
 
 # Run nodemon as a watcher to recompile Go
-npx nodemon main.go
+npx nodemon main.go &
```
davidlougheed, line 16: ah yes that would do it...

davidlougheed (approved):

davidlougheed:
looks good; maybe bump the version to 0.10.1 here as well?

PR #69, opened 2023-04-03
added types to overview

SanjeevLakhwani:


src/js/hooks.ts
v-rocheleau, line 1: Didn't know about this, pretty cool. 
We could add a comment linking to the react-redux doc [section](https://react-redux.js.org/using-react-redux/usage-with-typescript#define-typed-hooks) for this, since we will have to get acquainted with typed hooks once we merge.

src/js/components/Overview/PublicOverview.tsx
```diff
@@ -34,7 +34,7 @@ const PublicOverview = () => {
             <Counts />
             {sections.map(({ sectionTitle, charts }, i) => (
               <div key={i} className="overview">
-                <OverviewSection index={i} title={sectionTitle} chartData={charts} />
```
v-rocheleau, line 37: Why remove index here?

v-rocheleau (changes requested):

src/js/components/Overview/PublicOverview.tsx
SanjeevLakhwani: OverviewSection doesn't use index

v-rocheleau (approved):

PR #70, opened 2023-04-03
fix(bento-charts): version bump to fix pie charts labels position

v-rocheleau:
bento-charts 2.0.0 => 2.0.1


v-rocheleau:
Will test the pr image in bentov2 before merging

PR #71, opened 2023-04-05
added types to provenance

SanjeevLakhwani:


src/js/hooks.ts
```diff
@@ -5,3 +5,22 @@ import type { RootState, AppDispatch } from './store';
 // refer to https://react-redux.js.org/using-react-redux/usage-with-typescript#define-typed-hooks for more info
 export const useAppDispatch: () => AppDispatch = useDispatch;
 export const useAppSelector: TypedUseSelectorHook<RootState> = useSelector;
+
+// ################### TRANSLATION HOOKS ###################
+import { useTranslation } from 'react-i18next';
+import { DEFAULT_TRANSLATION, NON_DEFAULT_TRANSLATION } from '@/constants/configConstants';
+import { NamespaceTranslationFunction } from '@/types/translation';
```
v-rocheleau, line 12: Place imports at the top

v-rocheleau (approved):
Lgtm, just a little import nitpick

src/js/hooks.ts
SanjeevLakhwani: it is to separate both, in future when we get more hooks they would end up going into a new folder and separate files

PR #72, opened 2023-04-05
Feature/about_translation

noctillion:
Description:
This PR adds support for translating the `about.html` content in the Overview tab. The changes introduced include:

Changes:
- Basic templates eng_about.html and fr_about.html
- Fetching both English and French versions of `x_about.html` content using async thunk.
- Dynamically rendering the `x_about.html` content in the Overview tab based on the active language.

### DIFF
--- a/src/js/constants/contentConstants.js
+++ b/src/js/constants/contentConstants.js
@@ -1,1 +1,1 @@
-export const aboutUrl = '/public/about.html';
+export const aboutUrl = '/public/';
--- a/src/js/constants/contentConstants.js
+++ b/src/js/constants/contentConstants.js
@@ -1,1 +1,1 @@
-export const aboutUrl = '/public/';
+export const partialAboutUrl = '/public/';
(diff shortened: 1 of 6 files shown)


src/js/constants/contentConstants.js
```diff
@@ -1 +1 @@
-export const aboutUrl = '/public/about.html';
+export const aboutUrl = '/public/';
```
brouillette, line 1: Nitpick: 'might be worth renaming `aboutUrl` to `partialAboutUrl` for consistency

noctillion: Good idea!

noctillion:
Thanks for the review!

SanjeevLakhwani (changes requested):
I would suggest using useTranslation hook for it as it would abstract the language change and a requirement for a french file, which is currently being called on initial access. 

If working with a work around, other changes would be to use two constants for en and fr files.

Also, this would cause a merge conflict as bento-public is now in typescript

noctillion:
Closed to change to PR- Feature/about translation ts #76

PR #76, opened 2023-05-03
Feature/about translation ts

noctillion:
Description:
This PR adds support for translating the about.html content in the Overview tab. The changes introduced include:

Changes:

Basic templates eng_about.html and fr_about.html
Fetching both English and French versions of x_about.html content using async thunk.
Dynamically rendering the x_about.html content in the Overview tab based on the active language.

### DIFF
--- a/src/js/constants/contentConstants.ts
+++ b/src/js/constants/contentConstants.ts
@@ -1,1 +1,1 @@
-export const partialAboutUrl = '/public/';
+export const partialAboutUrl = '/public';
--- a/src/js/constants/contentConstants.ts
+++ b/src/js/constants/contentConstants.ts
@@ -1,1 +1,1 @@
-export const aboutUrl = '/public/about.html';
+export const partialAboutUrl = '/public/';
--- a/src/js/features/content/content.store.ts
+++ b/src/js/features/content/content.store.ts
@@ -39,8 +39,7 @@ const content = createSlice({
       state.isFetchingAbout = true;
     });
     builder.addCase(makeGetAboutRequest.fulfilled, (state, { payload }: PayloadAction<{ en: string; fr: string }>) => {
-      state.about.en = payload.en;
-      state.about.fr = payload.fr;
+      state.about = { ...payload}
       state.isFetchingAbout = false;
     });
     builder.addCase(makeGetAboutRequest.rejected, (state) => {
--- a/src/js/features/content/content.store.ts
+++ b/src/js/features/content/content.store.ts
@@ -4,29 +4,30 @@ import { partialAboutUrl } from '@/constants/contentConstants';
 import { printAPIError } from '@/utils/error.util';
 
 export const makeGetAboutRequest = createAsyncThunk('content/getAboutHTML', async () => {
-  const en_aboutHTML = await axios
-    .get(`${partialAboutUrl}en_about.html`)
+  const en = await axios
+    .get(`${partialAboutUrl}/en_about.html`)
     .then((res) => res.data)
     .catch(printAPIError);
 
-  const fr_aboutHTML = await axios
-    .get(`${partialAboutUrl}fr_about.html`)
+  const fr = await axios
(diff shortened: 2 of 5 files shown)


src/js/constants/contentConstants.ts
```diff
@@ -1 +1 @@
-export const aboutUrl = '/public/about.html';
+export const partialAboutUrl = '/public/';
```
v-rocheleau, line 1: Replace `'/public/'` with `'/public'`. All of our url constants have no trailling slash.

src/js/features/content/content.store.ts
```diff
@@ -1,25 +1,32 @@
...
-      .catch(printAPIError(rejectWithValue))
-);
+export const makeGetAboutRequest = createAsyncThunk('content/getAboutHTML', async () => {
+  const en_aboutHTML = await axios
+    .get(`${partialAboutUrl}en_about.html`)
+    .then((res) => res.data)
+    .catch(printAPIError);
+
+  const fr_aboutHTML = await axios
+    .get(`${partialAboutUrl}fr_about.html`)
```
v-rocheleau, line 13: Add `/` before file name once partialAboutUrl has no trailling slash

src/js/features/content/content.store.ts
```diff
@@ -1,25 +1,32 @@
...
+
+  return { en_aboutHTML, fr_aboutHTML };
+});
 
 export type ContentState = {
   isFetchingAbout: boolean;
-  aboutHTML: string;
+  en_aboutHTML: string;
+  fr_aboutHTML: string;
 };
```
v-rocheleau, line 24: I would change the ContentState shape for:
```typescript
export type ContentState = {
  isFetchingAbout: boolean;
  about: {[key: string]: string}
};

const initialState: ContentState = {
  isFetchingAbout: true,
  about: {
    en: "",
    fr: ""
  }
};
```
This way, we can get the about content by key and avoid the if blocks in PublicOverview's useEffect:
```typescript
  const { isFetchingAbout, about } = useAppSelector((state) => state.content);

  const [aboutContent, setAboutContent] = useState('');

  useEffect(() => {
    const activeLanguage = i18n.language;
    const activeAbout = about[activeLanguage];
    setAboutContent(activeAbout);
  }, [i18n.language, about]);
```
Like this we won't have to change the PublicOverview component if we add new languages

v-rocheleau (changes requested):

noctillion:
Thanks for the review and excellent suggestions!

src/js/features/content/content.store.ts
```diff
@@ -30,8 +38,9 @@ const content = createSlice({
     builder.addCase(makeGetAboutRequest.pending, (state) => {
       state.isFetchingAbout = true;
     });
-    builder.addCase(makeGetAboutRequest.fulfilled, (state, { payload }: PayloadAction<string>) => {
-      state.aboutHTML = payload;
+    builder.addCase(makeGetAboutRequest.fulfilled, (state, { payload }: PayloadAction<{ en: string; fr: string }>) => {
+      state.about.en = payload.en;
+      state.about.fr = payload.fr;
```
v-rocheleau, line 43: State update can be simplified with:
```typescript
state.about = {...payload};
```

v-rocheleau (changes requested):

src/js/features/content/content.store.ts
```diff
@@ -39,8 +39,7 @@ const content = createSlice({
       state.isFetchingAbout = true;
     });
     builder.addCase(makeGetAboutRequest.fulfilled, (state, { payload }: PayloadAction<{ en: string; fr: string }>) => {
-      state.about.en = payload.en;
-      state.about.fr = payload.fr;
+      state.about = { ...payload}
```
v-rocheleau, line 42: Missing semicolon

v-rocheleau (approved):

PR #77, opened 2023-05-03
Fixed rendering of charts in search response

SanjeevLakhwani:
- Removed unnecessary is valid variable
- The typescript interface for search response was incorrect causing issues for search not being loaded correctly which is now resolved


davidlougheed (approved):
lgtm

PR #81, opened 2023-05-17
add beacon UI

gsfk:
First pass at a beacon UI for bento_public:
- conditionally renders a "Beacon" tab when Bento env variable `BENTO_BEACON_UI_ENABLED` is true
- can query variants, metadata or both
- metadata queries use the same katsu config settings as traditional bento_public search
- metadata can be queried even when variants are absent
- currently does "happy path" behaviour only, with some form verification.

To come (in Bento 13):
- better user guidance / feedback / unhappy path
- Typescript annnotations

Requirements: 
- Bento 12
- [Beacon 0.6.0](https://github.com/bento-platform/bento_beacon/releases/tag/v0.6.0)
- latest Katsu develop ([relevant code](https://github.com/bento-platform/katsu/pull/402) is merged but not yet released)


src/js/components/Beacon/BeaconQueryUi.js
```diff
@@ -98,47 +92,73 @@ const BeaconQueryUi = () => {
...
-    marginTop: '25px',
-    marginLeft: '50px',
-  };
-
-  const buttonStyle = { margin: '10px 0' };
-
-  const buttonAreaStyle = {
-    display: 'flex',
-    padding: 0,
+    justfifyContent: "space-between"
```
brouillette: nitpick: typo "justfify"

gsfk: Nice, thanks. I **_love_** how css fails sliently.

src/js/components/Beacon/BeaconQueryUi.js
```diff
@@ -0,0 +1,215 @@
...
+  ));
+  const formInitialValues = { 'Assembly ID': beaconAssemblyIds.length == 1 && beaconAssemblyIds[0] };
+
+  // beacon request handling
+
+  const packageFilters = (values) => {
+
+    // ignore optional first filter when left blank 
+    if (filters.length == 1 && !values.filterId1) {
+      return []
```
davidlougheed, line 63: semi (we should add a linter rule for consistency)

src/js/components/Beacon/BeaconQueryUi.js
```diff
@@ -0,0 +1,215 @@
...
+
+  // following GA4GH recommendations, UI is one-based, but API is zero-based, "half-open"
+  // so to convert to zero-based, we only modify the start value 
+  // see eg https://genome-blog.soe.ucsc.edu/blog/2016/12/12/the-ucsc-genome-browser-coordinate-counting-systems/ 
+  const convertStartToZeroBased = (start) => Number(start)-1 
+
+  const packageBeaconJSON = (values) => {
+    let query = {};
+    const payloadFilters = packageFilters(values);
+    const hasVariantsQuery = (values && values['Chromosome']) || values['Variant start'] || values['Reference base(s)'];
```
davidlougheed, line 89: instead of `(values && values['Chromosome'])`, use `values?.['Chromosome']`

src/js/components/Beacon/BeaconQueryUi.js
```diff
@@ -0,0 +1,215 @@
...
+    return (
+      <Tooltip title={text}>
+        <InfoCircleOutlined />
+      </Tooltip>
+    );
+  };
+  
+  // styles
+
+  const wrapperStyle = {
```
davidlougheed, line 139: move style constants outside function to avoid redefining every render

src/js/components/Beacon/BeaconSearchResults.js
```diff
@@ -0,0 +1,39 @@
...
+  const biosamples = info?.bento?.biosamples ?? {};
+  const biosampleCount = biosamples.count;
+  const biosampleChartData = serializeChartData(biosamples.sampled_tissue ?? []);
+
+  const experiments = info?.bento?.experiments ?? {};
+  const experimentCount = experiments.count ?? 0;
+  const experimentChartData = serializeChartData(experiments?.experiment_type ?? []);
+
+  // shown when count = 0
+  const message = 'Insufficient data available.';
```
davidlougheed, line 23: give an all caps constant style name and move outside function

src/js/components/Beacon/Filter.js
```diff
@@ -0,0 +1,67 @@
...
+  }, [valueOptions]);
+
+  const renderLabel = (searchField) => {
+    const units = searchField.config?.units;
+    const unitsString = units ? ` (${units})` : '';
+    return searchField.title + unitsString;
+  };
+
+  const searchKeyOptions = (arr) => {
+    return arr.map((qs) => ({
```
davidlougheed, line 33: can be an expression-style function

src/js/components/Beacon/Filter.js
```diff
@@ -0,0 +1,67 @@
...
+      options: qs.fields.map((field) => ({
+        label: renderLabel(field),
+        value: field.id,
+        optionsThisKey: searchValueOptions(field.options),
+      })),
+    }));
+  };
+
+  const searchValueOptions = (arr) => {
+    return arr.map((v) => ({ label: v, value: v }));
```
davidlougheed, line 44: can be an expression-style function

src/js/components/Beacon/Filters.js
```diff
@@ -0,0 +1,51 @@
...
+  const hasMaxFilters = activeFilters.length >= maxFilters;
+
+  const handleAddFilter = () => {
+    const filterIndex = filters.length + 1;
+    const f = newFilter(filterIndex);
+    setFilters((filters) => [...filters, f]);
+  };
+
+  // UI starts with an optional filter, which can be left blank
+  const isRequired = filters.length > 1
```
davidlougheed, line 29: semi

src/js/components/Beacon/VariantsForm.js
```diff
@@ -0,0 +1,108 @@
...
+// TODOs:
+// show which fields are required without removing the ability to leave the form blank
+
+// form state has to be one of these:
+// empty (except for autofilled assemblyID)
+// chrom, start, assemblyID, end
+// chrom, start, assemblyID, ref, alt
+
+const VariantsForm = ({ assemblyIdOptions }) => {
+  const formStyle = {
```
davidlougheed, line 13: move outside (same goes for a bunch of vars below)

src/js/components/Beacon/VariantsForm.js
```diff
@@ -0,0 +1,108 @@
...
+      placeholder: 'A, C, G, T or N',
+      initialValue: '',
+      type: 'string',
+    },
+    assemblyId: { name: 'Assembly ID', rules: [{}], placeholder: '', initialValue: '' },
+  };
+
+  const formItemStyle = { width: '150px', padding: '0 5px 0 0px' };
+
+  const VariantInput = ({ field }) => {
```
davidlougheed, line 70: move outside

src/js/components/Beacon/VariantsForm.js
```diff
@@ -0,0 +1,108 @@
...
+    return (
+      <div style={formItemStyle}>
+        <Form.Item name={field.name} label={field.name} rules={field.rules} help={field.help}>
+          <Input placeholder={field.placeholder} />
+        </Form.Item>
+      </div>
+    );
+  };
+
+  const AssemblyIdSelect = ({ field, options }) => {
```
davidlougheed, line 80: move outside

src/js/components/Search/SearchResults.tsx
```diff
@@ -1,84 +1,32 @@
...
   const individualCount = useAppSelector((state) => state.query.individualCount);
   const message = useAppSelector((state) => state.query.message);
 
-  const wrapperStyle = {
-    padding: '40px',
-    minHeight: '150px',
-    maxHeight: '475px',
-  };
+  // existing search code has message existing as only sign of insufficent data
+  const hasInsufficientData = Boolean(message)
```
davidlougheed, line 15: equality might be clearer, Boolean() is not a common pattern in JS - also missing semi

src/js/components/TabbedDashboard.tsx
```diff
@@ -26,15 +28,18 @@ const TabbedDashboard = () => {
...
     dispatch(makeGetDataRequestThunk());
     dispatch(makeGetSearchFields());
     dispatch(makeGetProvenanceRequest());
+
   }, []);
 
   const isFetchingOverviewData = useAppSelector((state) => state.data.isFetchingData);
   const isFetchingSearchFields = useAppSelector((state) => state.query.isFetchingFields);
+  const isFetchingBeaconConfig = useAppSelector((state) => state.beaconConfig?.isFetchingBeaconConfig)
+  const renderBeaconUi = useAppSelector((state) => state.config?.beaconEnabled)
```
davidlougheed, line 42: semis

davidlougheed (changes requested):
a few notes. also, in the future, we should not write anything for bento public in JS at all anymore - it should be TS from the start (and I think starting at 13 we should enforce TS-only).

src/js/components/Search/SearchResults.tsx
gsfk: I'll just remove this variable entirely if you think implicit is preferable: ie use `hasInsufficientData={message}` in the call to the component

src/js/components/Beacon/BeaconQueryUi.js
gsfk: Moved out. There's _so much_ style here and elsewhere that it would probably work better in a traditional css file, but it won't match the rest of the project.

src/js/components/Search/SearchResults.tsx
gsfk: It will have to be `message !== undefined` since the current expected value of "message" is either undefined, or the string "Insufficient data available." This is the nature of the katsu response, which could always be improved. 

I'll throw that in if you can live with it.

davidlougheed: I think that sounds good!

gsfk: Looking further into the redux it seems like "message" is given a default value of empty string if it's missing from the katsu response, so  `message !== ""` should be enough for now.

src/js/features/beacon/beaconConfig.js
```diff
@@ -0,0 +1,36 @@
+import { createSlice, createAsyncThunk } from '@reduxjs/toolkit';
```
davidlougheed, line 1: this file doesn't follow the naming scheme for other public features

src/js/features/beacon/beaconQuery.js
```diff
@@ -0,0 +1,37 @@
+import { createSlice, createAsyncThunk } from '@reduxjs/toolkit';
```
davidlougheed, line 1: this file doesn't follow the naming scheme followed by other public features

davidlougheed (changes requested):

src/js/features/beacon/beaconConfig.js
```diff
@@ -0,0 +1,36 @@
...
+import axios from 'axios';
+import { printAPIError } from '../../utils/error.util.ts';
+
+export const getBeaconConfig = createAsyncThunk('beaconConfig/getBeaconConfig', async (_, { getState }) => {
+  const beaconInfoEndpoint = getState()?.config?.beaconUrl + '/info';
```
davidlougheed, line 6: this function isn't actually async since there are no awaits inside it

src/js/features/beacon/beaconConfig.js
gsfk: It sure did when I wrote the code! Pretty much all of the beacon redux is just  modelled after what was in the codebase at the time. 

The redux code has changed quite a bit since then, beyond just the naming scheme. One of the tasks I've left for myself is to align the beacon redux with the current redux elsewhere. I'm not convinced this needs to be done immediately.

src/js/features/beacon/beaconQuery.js
gsfk: as above

src/js/features/beacon/beaconConfig.js
gsfk: No trouble to remove the async here. We probably don't need `createAsyncThunk` then, although this same code pattern is used throughout.

davidlougheed: it's still returning a promise, so probably still need createAsyncThunk - it's just that you don't await the promise, so the function doesn't need to be async

src/js/components/Beacon/BeaconQueryUi.js
```diff
@@ -0,0 +1,205 @@
...
+};
+
+const BUTTON_AREA_STYLE = {
+  padding: '20px 5px',
+};
+
+const BUTTON_STYLE = {
+  margin: '0 10px 0 0',
+};
+
```
SanjeevLakhwani, line 55: I think it might be better to move the styles to a constants folder

src/js/components/Beacon/BeaconQueryUi.js
```diff
@@ -0,0 +1,205 @@
...
+    if (!beaconUrl) {
+      return;
+    }
+    launchEmptyQuery();
+  }, [config]);
+
+  // set assembly id options mataching what's in gohan
+  useEffect(() => {
+    form.setFieldsValue(formInitialValues);
+  }, [config]);
```
SanjeevLakhwani, line 79: cant the 2 be combined given the similar dependencies

src/js/components/Beacon/Filter.js
```diff
@@ -0,0 +1,64 @@
...
+        value: field.id,
+        optionsThisKey: searchValueOptions(field.options),
+      })),
+    }));
+
+  const searchValueOptions = (arr) => arr.map((v) => ({ label: v, value: v }));
+
+  return (
+    <Space.Compact>
+      <Form.Item name={`filterId${filter.index}`} rules={[{ required: isRequired, message: 'search field required' }]}>
```
SanjeevLakhwani, line 46: could extract Item from form

src/js/components/Beacon/Filters.js
```diff
@@ -0,0 +1,58 @@
...
+    const f = newFilter(filterIndex);
+    setFilters((filters) => [...filters, f]);
+  };
+
+  // UI starts with an optional filter, which can be left blank
+  const isRequired = filters.length > 1;
+
+  return (
+    <Form.Item>
+      <div style={{ display: 'flex', padding: 0 }}>
```
SanjeevLakhwani, line 35: Can replace div with space. The whole of bento public does that. It replaces div with space & react fragment

src/js/components/Beacon/VariantsForm.js
```diff
@@ -0,0 +1,90 @@
...
+// chrom, start, assemblyID, end
+// chrom, start, assemblyID, ref, alt
+
+// forgiving chromosome regex
+// accepts X, Y, etc and any one- or two-digit non-zero number
+// note that, eg, polar bears have 37 pairs of chromosomes...
+const CHROMOSOME_REGEX = /^([1-9][0-9]?|X|x|Y|y|M|m|MT|mt)$/;
+
+const NUCLEOTIDES_REGEX = /^(a|c|g|t|n|A|C|G|T|N)*$/;
+const DIGITS_REGEX = /^[0-9]+$/;
```
SanjeevLakhwani, line 20: constant file worthy

src/js/components/Beacon/VariantsForm.js
```diff
@@ -0,0 +1,90 @@
...
+
+const VariantsForm = ({ assemblyIdOptions }) => {
+  const formStyle = {
+    maxWidth: '1200px',
+    display: 'flex',
+    background: 'white',
+    padding: '10px',
+    borderRadius: '10px',
+  };
+
```
SanjeevLakhwani, line 71: constant file worthy

SanjeevLakhwani (commented):
Great work. Also would be great to know what would be the timeline for it to get into typescript.

src/js/components/Beacon/BeaconQueryUi.js
gsfk: not a css file?

main.go
```diff
@@ -25,17 +25,21 @@ const ConfigLogTemplate = `Config --
...
+	ServiceId          string `envconfig:"BENTO_PUBLIC_SERVICE_ID"`
+	PackageJsonPath    string `envconfig:"BENTO_PUBLIC_PACKAGE_JSON_PATH" default:"./package.json"`
+	StaticFilesPath    string `envconfig:"BENTO_PUBLIC_STATIC_FILES_PATH" default:"./www"`
+	ClientName         string `envconfig:"BENTO_PUBLIC_CLIENT_NAME"`
+	KatsuUrl           string `envconfig:"BENTO_PUBLIC_KATSU_URL"`
+	BentoPortalUrl     string `envconfig:"BENTO_PUBLIC_PORTAL_URL"`
+	Port               int    `envconfig:"INTERNAL_PORT" default:"8090"`
+	Translated         bool   `envconfig:"BENTO_PUBLIC_TRANSLATED" default:"true"`
+	BeaconUrl          string `envconfig:"BEACON_URL"`
+	BeaconUiEnabled	   bool   `envconfig:"BENTO_BEACON_UI_ENABLED"`
```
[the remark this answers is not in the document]
gsfk: These are both available in the container environment, see [here](https://github.com/bento-platform/bento/blob/releases/v2.12/lib/public/docker-compose.public.yaml#L16-L17).  In fact it looks like a lot of example.sever.env is deprecated.

src/js/components/Beacon/Filter.js
gsfk: Removing it outright will remove both `name` and `rules`. Then what?

src/js/components/Beacon/VariantsForm.js
gsfk: Which part? This is a component.

davidlougheed: only if re-used, otherwise it doesnt make sense to have component-specific styles outside the component

davidlougheed: however, it should be outside of the function here

src/js/components/Beacon/BeaconQueryUi.js
gsfk: Yes, good catch. Judging from the comments, the first one started out with an empty dependency array, and I guess I changed it without noticing.

src/js/components/Beacon/VariantsForm.js
gsfk: I'm lukewarm on this one. It's three lines used only in this component.

src/js/components/Beacon/BeaconQueryUi.js
SanjeevLakhwani: I am gonna look into this I get the point that css is not reused but having css in component doesnt help in understanding a component, the layout + functionality is the main thing of a component and style though important is secondary for a jsx file. Do as you like, I shall look into a standard for this.

src/js/components/Beacon/Filter.js
SanjeevLakhwani: I do not mind it, I just prefer Item by itself when it is used repetitively but also in this case, Item is not within a form and Item is not clear by itself so this works.

davidlougheed (approved):
lgtm

PR #82, opened 2023-05-25
fix: language change with url param

v-rocheleau:
**Issues**: After opening the site, we cant change the language until a route change is done

**Causes**: 
- In `src/js/index.tsx` the language url param was being set before i18n was initialized, causing it to be undefined (`/#/undefined`) until the next route change (when i18n.language is `en` or `fr`).
- Missing trailing `/` after language url param was preventing path replacement in `SiteHeader.changeLanguage`

**Other:**
- `SiteHeader.tsx` now accessing the i18n instance with the useTranslation hook
- added `www/` dir to .gitignore

### DIFF
--- a/src/js/index.tsx
+++ b/src/js/index.tsx
@@ -5,6 +5,7 @@ import { useTranslation } from 'react-i18next';
 import { HashRouter, Routes, Route, useParams, useNavigate } from 'react-router-dom';
 import { Layout } from 'antd';
 import { ChartConfigProvider } from 'bento-charts';
+import { SUPPORTED_LNGS } from './constants/configConstants';
 
 import './i18n';
 import '../styles.css';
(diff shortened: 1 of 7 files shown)


src/js/index.tsx
```diff
@@ -25,10 +25,10 @@ const App = () => {
   useEffect(() => {
     if (lang === 'en' || lang === 'fr') {
       i18n.changeLanguage(lang);
-    } else {
-      navigate(`/${i18n.language}`);
+    } else if (i18n.language) {
+      navigate(`/${i18n.language}/`);
     }
```
davidlougheed, line 30: should this have a fallback else branch to navigate to `/en/`? otherwise this looks good to me

davidlougheed (approved):
lgtm

PR #83, opened 2023-05-30
Add new lint rule & run Prettier check in CI

davidlougheed:
Also formats the whole project according to Prettier


v-rocheleau (approved):
lgtm

PR #85, opened 2023-06-08
style: responsive beacon UI + search results pane

davidlougheed:


gsfk:
Beacon page looks great, seems like "loading" is broken on Search tab though:

search

davidlougheed:
oops, thanks, i'll take a look

PR #89, opened 2023-06-14
Translated beacon UI

SanjeevLakhwani:


src/js/components/Beacon/VariantsForm.js
```diff
@@ -19,52 +21,54 @@ const CHROMOSOME_REGEX = /^([1-9][0-9]?|X|x|Y|y|M|m|MT|mt)$/;
...
 const FORM_STYLE = {
   display: 'flex',
   flexDirection: 'column',
 };
 const FORM_ROW_GUTTER = [12, 0];
 
 const VariantsForm = ({ assemblyIdOptions }) => {
+  const { t: td } = useTranslation(DEFAULT_TRANSLATION);
+
+  const FORM_FIELDS = {
```
davidlougheed, line 33: usememo + dependency on td

davidlougheed (changes requested):

src/js/components/Beacon/VariantsForm.js
SanjeevLakhwani: good point done

src/js/components/Beacon/VariantsForm.js
```diff
@@ -19,52 +21,57 @@ const CHROMOSOME_REGEX = /^([1-9][0-9]?|X|x|Y|y|M|m|MT|mt)$/;
...
+      },
+      referenceBases: {
+        name: td('Reference base(s)'),
+        rules: [{ pattern: NUCLEOTIDES_REGEX, message: td('enter any combination of A, C, G, T, or N') }],
+        placeholder: 'A, C, G, G, T or N',
+        initialValue: '',
+        type: 'string',
+      },
+      alternateBases: {
+        name: td('Alternate base(s)'),
```
gsfk, line 64: Translating the `name` property here breaks variant search, since form information is filled in according to the name. Instead of translating `name`, add a new field called "label" and translate that, eg:

```
name: 'Chromosome,
label: td('Chromosome')
```
and then update the label field in `VariantInput` and `AssemblyIdSelect` so that it says `label={field.label}` instead of the current `label={field.name}`. Sorry if I missed this before.

gsfk (changes requested):
Filters seem to work now, has a variant search bug that I missed the first time.

src/js/components/Beacon/VariantsForm.js
SanjeevLakhwani: done, nw. I translated the label in the final

gsfk: ah, this is a cleaner solution, nice.

gsfk (commented):
Okay, variant search works now, but units are not translated: 

ans

SanjeevLakhwani:
> Okay, variant search works now, but units are not translated:
> 
> ans

The units are translated, a translation needs to exist for them to have a french version

src/js/components/Beacon/BeaconQueryUi.js
```diff
@@ -55,7 +60,7 @@ const BeaconQueryUi = () => {
 
   const assemblyIdOptions = beaconAssemblyIds.map((assembly) => (
     <Select.Option key={assembly} value={assembly}>
-      {assembly}
+      {t(assembly)}
```
davidlougheed, line 63: this will likely never have a translation

davidlougheed (approved):
overall looks good - will need typescript reconciliation

src/js/components/Beacon/BeaconQueryUi.js
SanjeevLakhwani: I do not know, but doesn't hurt if any one would ever want to.

PR #90, opened 2023-06-14
Added onClick functionality to charts

SanjeevLakhwani:
clicking a chart section to charts now redirects you to corresponding search

### DIFF
--- a/src/js/components/Overview/Chart.tsx
+++ b/src/js/components/Overview/Chart.tsx
@@ -5,9 +5,12 @@ import { BarChart, PieChart } from 'bento-charts';
 
 import { CHART_HEIGHT } from '@/constants/overviewConstants';
 import { ChartData } from '@/types/data';
+import { useNavigate, useLocation } from 'react-router-dom';
 
-const Chart = ({ chartType, data, units }: ChartProps) => {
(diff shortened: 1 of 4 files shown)


src/js/components/Overview/Chart.tsx
```diff
@@ -22,10 +25,23 @@ const Chart = ({ chartType, data, units }: ChartProps) => {
...
       case chartTypes.PIE:
-        return <PieChart data={data} height={CHART_HEIGHT} preFilter={removeMissing} dataMap={translateMap} />;
+        return (
+          <PieChart
+            data={data}
+            height={CHART_HEIGHT}
+            preFilter={removeMissing}
+            dataMap={translateMap}
+            onClick={(d) => {
+              navigate(`${location.pathname.slice(0, 3)}/search?${id}=${d.name}`);
```
davidlougheed, line 41: i think it makes more sense to inject the language directly here rather than slicing it from the location - this is less immediately obvious

davidlougheed (changes requested):

davidlougheed (approved):
lgtm - might be good to make a custom hook eventually to inject the language prefix into a URL; that way we don't always have to write the `/${i18n.language}` part

PR #96, opened 2023-07-03
fix: error when selecting query option + missing URL params in some cases

davidlougheed:
- fix: error when selecting query option + bad selectors
- fix: include query parameters when returning to search page
- refact: move search URL-handling logic to a single location


gsfk (changes requested):
Now launches every query twice: two identical network calls, and two flashes of the interface, although only one flash for some (cached?) queries.

gsfk (approved):
Working as expected now for me.

PR #97, opened 2023-07-05
fix: chart click nav

v-rocheleau:
Chart click navigation was failing due to a faulty query parameters check (Search.tsx checkQueryParamsEqual).
Given the parameter "age", the function would spread each letter as its own parameter ("a", "g", "e"), causing `undefined === undefined` which evaluates to true always.


davidlougheed (approved):
lgtm

PR #100, opened 2023-07-26
fix: api error catching in beacon feature

davidlougheed:


gsfk (approved):
Ah thanks... I swear I was going to start error handling code...

PR #101, opened 2023-08-02
Features/beacon hide variants search

gsfk:
First pass at conditional rendering of variants search form in Beacon tab:

- if there are no variants in the beacon, hide the variants form
- if variants are expected but gohan is down, returning errors, or claims there are no variants, show the variants form but disable it. 

(Variants are expected when beacon_config.json has param `useGohan=true`)


SanjeevLakhwani (approved):
lgtm

PR #103, opened 2023-09-06
[14] fix: overview sections alignment when no data

v-rocheleau:


src/js/components/Overview/LastIngestion.tsx
```diff
@@ -32,9 +34,13 @@ const LastIngestionInfo: React.FC = () => {
   );
 
   const hasData = Object.keys(lastEndTimesByDataType).length > 0;
+  const hasSections = sections.length > 0;
 
   return (
-    <Space direction="vertical" size={0}>
+    <ConditionalWrapper
+      condition={hasData && hasSections}
+      wrapper={children => <Space direction="vertical" size={0}>{children}</Space>}
```
davidlougheed, line 42: overall this looks good - in the future this might be better as a component class vs a function, since as a function this will always force a re-render of  ConditionalWrapper even if it didnt change but other parts of LastIngestionInfo did

vs if you made a component VerticalSpaceWrapper you could use it for the charts sections too + pass it in and it wouldnt change every render and use the component directly in ConditionalWrapper

davidlougheed (approved):

src/js/components/Overview/LastIngestion.tsx
v-rocheleau: Good point

davidlougheed (approved):
lgtm

PR #104, opened 2023-09-06
feat: data types labels display in overview

v-rocheleau:


davidlougheed (changes requested):
this should be `feat` not `perf`

i would rather these werent hard-coded, since phenopacket label especially needs to be explicitly changeable for projects like bearwatch and genovalia

you can now get all data types from `service-registry` via the `/data-types` endpoint, so this should provide an easier way to aggregate labels in bento_public

src/js/features/dataTypes/datatypes.store.ts
```diff
@@ -0,0 +1,39 @@
+import { PayloadAction, createAsyncThunk, createSlice } from '@reduxjs/toolkit';
+import axios from 'axios';
+
+// TODO: find a way to allow this without an auth token
+export const makeGetDataTypes = createAsyncThunk('datatypes/makeGetDataTypes', async () => {
```
davidlougheed, line 5: rename to `dataTypes/makeGetDataTypes` for consistency

davidlougheed (changes requested):

src/js/types/datatypes.ts
```diff
@@ -0,0 +1,24 @@
+export enum DataTypes {
```
davidlougheed, line 1: rename this one too

davidlougheed: (the file)

davidlougheed (approved):
lgtm as a temporary measure

PR #108, opened 2023-09-12
chore: update provenance styling + minor refactor

davidlougheed:
* make all tables the same size & give them borders
* render taxonomies in italics in the "Is about" section


noctillion (approved):
It works as expected! Lgtm.

PR #109, opened 2023-09-25
fix: translate 'Remove this chart'

davidlougheed:


gsfk (approved):
:+1: 

... I'm just assuming "tableau" is correct for "chart", since the translation already contains "Gestion des tableaux".

PR #112, opened 2023-09-28
feat: add support for choropleth map & clean up chart types

davidlougheed:


src/js/types/chartConfig.ts
```diff
@@ -0,0 +1,30 @@
...
+export const CHART_TYPE_PIE = 'pie';
+export const CHART_TYPE_BAR = 'bar';
+export const CHART_TYPE_CHOROPLETH = 'choropleth';
+
+/*
+ChartConfig: represents what is stored in the configuration file for describing a chart, without any attached data or
+the field metadata (description / mapping information / units / etc.) By using a sum type here, we can optionally
+mandate configuration information for certain types of charts.
+ */
+export type ChartConfig =
```
v-rocheleau, line 13: Would make three distinct types and then sum type them like: 
```typescript
export type ChartConfig = PieChartConfig  | BarChartConfig | ChoroplethChartConfig;
```
Makes it more readable imo, and we may want to import a specific type later.

v-rocheleau (approved):
Lgtm, just a small comment on typing

src/js/types/chartConfig.ts
davidlougheed: I agree, did this already in another WIP branch.

PR #114, opened 2023-09-28
feat: allow for multi-column-width charts and improve responsivity

davidlougheed:
also stores widths in localstorage + redraws charts on width change

making the charts (other than map) behave nicely w/ width change is a separate thing for bento_charts...


gsfk (approved):
Appears to work as expected, strictly in terms of adjusting the parent container width. (I don't have any map data to try).

A few design notes: 

- pages with mixed widths are not filled optimally, since charts always appear in fixed order. I don't know if freeing up the order will make charts harder to manage or not. 
- it's a bit confusing that the width control isn't on the chart itself, although "manage charts" does seem like a reasonable place to put this
- we should probably add some width elements to the public config: 
  - it would be useful to have a configurable default width per chart, since we know in advance for some charts that they will look better wider
  - we might not want width controls on all charts (do pie charts need them?). Here I'm mostly thinking of the amount of clutter the controls add to "manage charts". If every chart has controls this won't really matter

davidlougheed:
@gsfk regarding point 3a, there is a configurable default width per chart in config now - although existing configs will not have them and default to `1`. good point w/ 3b though...

gsfk:
ah, I guess I only have an "old" config

PR #116, opened 2023-10-11
feat: improve provenance display across multiple facets

davidlougheed:


gsfk (approved):
Works well, no issues in testing. 
- doi / http links both working, although where an identifier exists, the title is already a clickable link. Do we need to have two links to the same thing in a single entry? 
- Looks better now without so much tag-style text layout, although some remains, and it's not clear if it was deliberate (eg "Funders" in Acknowledgements is still tag-style)

PR #117, opened 2023-10-11
feat: last ingestion time from data services

noctillion:
This PR remove the last data ingestion from WES and instead retrieve the information from gohan and katsu services /data-types


src/js/features/ingestion/lastIngestion.store.ts
```diff
@@ -0,0 +1,75 @@
...
+    builder.addCase(fetchKatsuData.pending, (state) => {
+      state.isFetchingData = true;
+    });
+    builder.addCase(fetchGohanData.pending, (state) => {
+      state.isFetchingData = true;
+    });
+    builder.addCase(fetchKatsuData.fulfilled, (state, { payload }: PayloadAction<DataResponseArray>) => {
+      const uniqueIds = new Set(state.dataTypes.map((data) => data.id));
+      const newData = payload.filter((data) => !uniqueIds.has(data.id));
+      state.dataTypes = [...state.dataTypes, ...newData];
```
v-rocheleau, line 58: Missing `state.isFetchingData = false;` I think.
Also since we have standard data-types responses in katsu and gohan we can factor out the reducer in a function for both `fetchKatsuData.fulfilled` and `fetchGohanData.fulfilled`:
```typescript
const reduceServiceDataTypes = (state, { payload }: PayloadAction<DataResponseArray>) => {
      const uniqueIds = new Set(state.dataTypes.map((data) => data.id));
      const newData = payload.filter((data) => !uniqueIds.has(data.id));
      state.dataTypes = [...state.dataTypes, ...newData];
      state.isFetchingData = false;
};

// in extraReducers func
builder.addCase(fetchKatsuData.fulfilled, reduceServiceDataTypes);
builder.addCase(fetchGohanData.fulfilled, reduceServiceDataTypes);
```

v-rocheleau (changes requested):
Looks good, small extract method refactoring suggestion

src/js/features/ingestion/lastIngestion.store.ts
noctillion: Sure! Thanks for the insigths!

v-rocheleau (approved):
lgtm!

v-rocheleau (approved):

davidlougheed (changes requested):
one slightly problematic thing here is that the counts exposed here are not censored by small cell count. for overall data types it doesn't matter too much, but maybe counts should be censored if you have the threshold in the Go server, or this field should just be deleted

main.go
```diff
@@ -382,6 +388,17 @@ func main() {
 		return c.String(http.StatusOK, string(data))
 	})
 
+	e.GET("/gohan/data-types", gohanRequestPublic)
+
+	e.GET("/katsu/data-types", func(c echo.Context) error {
```
davidlougheed, line 393: this also needs to filter out counts, no?

main.go
```diff
@@ -201,11 +193,27 @@ func main() {
 		return jsonFormattedData, nil
 	}
 
-	wesRequestWithDetailsAndPublic := func(c echo.Context) error {
-		qs := url.Values{}
-		qs.Add("with_details", "true")
-		qs.Add("public", "true")
-		return wesRequest("/runs", qs, c, jsonDeserialize)
+	gohanRequestPublic := func(c echo.Context) error {
```
davidlougheed, line 196: you could make this a generic function and pass in the gohan/katsu URL as an argument, since the logic and structure should be identical for both endpoints.

davidlougheed (changes requested):

main.go
noctillion: I thought was only for gohan. But sure!

davidlougheed: counts are in both responses to my knowledge

src/js/features/ingestion/lastIngestion.store.ts
```diff
@@ -0,0 +1,73 @@
...
+    builder.addCase(fetchGohanData.pending, (state) => {
+      state.isFetchingData = true;
+    });
+    builder.addCase(fetchKatsuData.fulfilled, reduceServiceDataTypes);
+    builder.addCase(fetchGohanData.fulfilled, reduceServiceDataTypes);
+    builder.addCase(fetchKatsuData.rejected, (state) => {
+      state.isFetchingData = false;
+    });
+    builder.addCase(fetchGohanData.rejected, (state) => {
+      state.isFetchingData = false;
```
davidlougheed, line 68: these conflict with one another since you have two possibly parallel requests using the same state. this should be two isFetching type variables

davidlougheed: (this comment applies to this whole extraReducers section)

src/js/types/lastIngestionDataTypeResponse.ts
```diff
@@ -0,0 +1,9 @@
+export interface LastIngestionDataTypeResponse {
+  count: number | null;
+  id: string;
+  label: string;
+  last_ingested: string | null;
+  queryable: boolean;
+}
+
+export type DataResponseArray = LastIngestionDataTypeResponse[];
```
davidlougheed, line 9: is this file used anywhere? `LastIngestionDataTypeResponse` is duplicate here - you already defined `DataTypeResponse` in the store

src/js/features/ingestion/lastIngestion.store.ts
```diff
@@ -0,0 +1,73 @@
...
+// Initialize the state
+const initialDataTypeState: DataTypeState = {
+  isFetchingData: false,
+  dataTypes: [],
+};
+
+const reduceServiceDataTypes = (state: DataTypeState, { payload }: PayloadAction<DataResponseArray>) => {
+  const uniqueIds = new Set(state.dataTypes.map((data: DataTypeResponse) => data.id));
+  const newData = payload.filter((data: DataTypeResponse) => !uniqueIds.has(data.id));
+  state.dataTypes = [...state.dataTypes, ...newData];
```
davidlougheed, line 46: this means that the data type array order will depend on which service returns first, and will flip around randomly based on this from the user's perspective. it makes more sense to make the data types in the store an object in my opinion, and set

```ts
state.dataTypes = {
    ...state.dataTypes, 
    ...Object.fromEntries(payload.map((data: DataTypeResponse) => [data.id, data])),
};
```

and then, when rendering, use `Object.entries` and apply sorting based on the label.

src/js/components/Overview/LastIngestion.tsx
```diff
@@ -6,10 +6,14 @@ import { useTranslation } from 'react-i18next';
...
 import { getDataTypeLabel } from '@/types/dataTypes';
+import { LastIngestionDataTypeResponse, DataResponseArray } from '@/types/lastIngestionDataTypeResponse';
 
 const LastIngestionInfo: React.FC = () => {
   const { t, i18n } = useTranslation(DEFAULT_TRANSLATION);
-  const lastEndTimesByDataType = useAppSelector((state) => state.ingestionData?.lastEndTimesByDataType) || {};
+
+  const lastEndTimesByDataType: DataResponseArray = useAppSelector((state) => state.lastIngestionData?.dataTypes) || [];
+
+  const queryableDataTypes = lastEndTimesByDataType.filter((item: LastIngestionDataTypeResponse) => item.queryable);
```
davidlougheed, line 16: could apply sorting here to make sure there's a consistent order based on label (see below comment)

davidlougheed (changes requested):

src/js/components/Overview/LastIngestion.tsx
noctillion: Sure! Nice catch; thanks for the insights!

src/js/components/Overview/LastIngestion.tsx
```diff
@@ -7,9 +7,18 @@ import { DEFAULT_TRANSLATION } from '@/constants/configConstants';
...
+import { LastIngestionDataTypeResponse } from '@/types/lastIngestionDataTypeResponse';
+
 const LastIngestionInfo: React.FC = () => {
   const { t, i18n } = useTranslation(DEFAULT_TRANSLATION);
-  const lastEndTimesByDataType = useAppSelector((state) => state.ingestionData?.lastEndTimesByDataType) || {};
+
+  const dataTypesObject = useAppSelector((state) => state.lastIngestionData?.dataTypes) || {};
+  const dataTypesArray = Object.values(dataTypesObject);
+
+  const sortedDataTypes = dataTypesArray.sort((a, b) => a.label.localeCompare(b.label));
```
davidlougheed, line 18: `.sort` mutates the original value in JS, so it doesn't make sense to have two variables assigned to the same memory location here - either use `.toSorted` for not-in-place or (my preference), combine the assignments since `dataTypesArray` isn't used anywhere else:

```js
const sortedDataTypes = Object.values(dataTypesObject).sort((a, b) => a.label.localeCompare(b.label));
 ```

File: src/js/features/ingestion/lastIngestion.store.ts [hunk 9]
```diff
@@ -0,0 +1,69 @@
...
+  dataTypes: {},
+};
+
+const reduceServiceDataTypes = (state: DataTypeState, { payload }: PayloadAction<LastIngestionDataTypeResponse[]>) => {
+  state.dataTypes = {
+    ...state.dataTypes,
+    ...Object.fromEntries(payload.map((data) => [data.id, data])),
+  };
+  state.isFetchingKatsuData = false;
+  state.isFetchingGohanData = false;
```
davidlougheed, line 43: this shouldn't set both to false since it is only setting data for one of the data types. you could move the assignment out of this function and make this function just handle the data type assignments, accepting `state` and `payload` as arguments, instead.

davidlougheed (changes requested):
almost there! a few little things left

davidlougheed (approved):
lgtm

PR #125, opened 2023-11-09
feat: beacon validation and errors

gsfk:
Form error checking and user feedback for Beacon search UI:
- block submission of incorrect forms
- show user feedback for errors in the form
- show user feedback for errors returned by beacon
- better tooltip helptext

Some new text requires translation. 

The sheer amount of instructions and feedback suggests that the form could probably be more intuitive to use. There is some discussion [here](https://github.com/bento-platform/bento_public/blob/features/beacon-validation-and-errors/src/js/components/Beacon/BeaconQueryUi.tsx#L145-L152) in the code.

### DIFF
--- a/src/js/components/Beacon/BeaconErrorMessage.tsx
+++ b/src/js/components/Beacon/BeaconErrorMessage.tsx
@@ -1,4 +1,4 @@
-import React, { Dispatch, SetStateAction } from 'react';
+import React from 'react';
 import { Alert } from 'antd';
 
 const BeaconErrorMessage = ({ message, onClose }: BeaconErrorMessageProps) => {


davidlougheed (changes requested):
preliminary comment + translations should be in place pre-merge

gsfk:
draft translation pr is [here](https://github.com/bento-platform/bento_public/pull/126), just needs French portions filled in.

src/js/components/Beacon/BeaconQueryUi.tsx
```diff
@@ -17,33 +17,61 @@ import {
...
 
-const VARIANTS_HELP =
-  'Variants search requires the fields Chromosome, Variant start, Assembly ID, and at least one of Variant end or Alternate base(s).';
-const METADATA_HELP = 'Search over clinical or phenotypic properties.';
 const STARTER_FILTER = { index: 1, active: true };
 
+// complexity of instructions suggests the form isn't intuitive enough
+const VARIANTS_INSTRUCTIONS_TITLE = 'Variant search';
+const VARIANTS_INSTRUCTIONS_LINE1 =
+  'To search for all variants inside a range: fill both "Variant start" and "Variant end", all variants inside the range will be returned. You can optionally filter by reference or alternate bases.';
```
davidlougheed, line 30: i guess bento_public linting doesn't enforce this, but i'd prefer a max 120-char line width here

src/js/components/Beacon/BeaconQueryUi.tsx
```diff
@@ -113,7 +141,46 @@ const BeaconQueryUi = () => {
...
 
+  const variantsFormValid = (formValues: FormValues) => {
+    // valid variant form states:
+    // empty except possibly autofilled assemblyID (no variant query)
+    // chrom, start, assemblyID, end (range query)
+    // chrom, start, assemblyID, alt, ref (sequence query)
+    // https://docs.genomebeacons.org/variant-queries/
+
+    // as an alternative, we could require "end" always, then form logic would be less convoluted
+    // just set start=end to search for SNPs
```
davidlougheed, line 152: when we eventually have chrom autofill from gohan/ref service, we could use the information in the ref service to auto-fill `end` to the end of the chromosome if it's unset (and change it if the user changes the chromosome to the length of the new chromosome if it === the previous chrom length) or something like this. for now, i think this is fine even though it's convoluted.

src/js/components/Beacon/BeaconQueryUi.tsx
```diff
@@ -122,17 +189,47 @@ const BeaconQueryUi = () => {
...
     form.resetFields();
     form.setFieldsValue(formInitialValues);
+    clearFormError();
+    setErrorAlertClosed(false);
     launchEmptyQuery();
   };
 
-  const SearchToolTip = ({ text }: { text: string }) => {
+  const handleValuesChange = (_: Partial<FormValues>, allValues: FormValues) => {
+    form.validateFields(['Chromosome', 'Variant start', 'Variant end', 'Reference base(s)', 'Alternate base(s)']);
```
davidlougheed, line 198: this list of fields is used once above too for `const empty` - maybe worth factoring out?

davidlougheed: (all fields, i guess?)

davidlougheed (commented):
overall looks good to me with a few minor code style things

src/js/components/Beacon/BeaconQueryUi.tsx
gsfk: Do people really want to search entire chromosomes?

davidlougheed: not really no, that's true

gsfk: could perhaps work for genes though I guess

src/js/components/Beacon/BeaconQueryUi.tsx
gsfk: yeah, there are a lot of variant form names lying around in the code (and not too helpfully the correct name for this collection is something like allVariantFormFieldsExceptAssemblyID... )

In `empty` we're checking each name individually, where in validate fields we're using all names together. So a possible refactor is something like:

`const empty  = !allVariantFormFieldsExceptAssemblyID.some(field => formValues[field])`

but i'm not sure this is better....

src/js/components/Beacon/BeaconQueryUi.tsx
```diff
@@ -1,11 +1,11 @@
...
-import { useTranslation } from 'react-i18next';
-import { Button, Card, Col, Form, Row, Tooltip } from 'antd';
+import React, { useEffect, useState, ReactNode } from 'react';
+import { useAppSelector, useAppDispatch, useTranslationDefault } from '@/hooks';
+import { Button, Card, Col, Form, Row, Space, Tooltip, Typography } from 'antd';
 import { InfoCircleOutlined } from '@ant-design/icons';
 import Filters from './Filters';
 import BeaconSearchResults from './BeaconSearchResults';
 import VariantsForm from './VariantsForm';
+import BeaconErrorMessage from './BeaconErrorMessage';
```
davidlougheed, line 8: group this with the other beacon component import

src/js/components/Beacon/BeaconErrorMessage.tsx
```diff
@@ -0,0 +1,17 @@
+import React, { Dispatch, SetStateAction } from 'react';
+import { Alert } from 'antd';
+
+const BeaconErrorMessage = ({ message, setErrorAlertClosed }: BeaconErrorMessageProps) => {
```
davidlougheed, line 4: I think a more generalizable parameter format would be to accept an `onClose` function, which is bound directly to the Alert - rather than passing a state setter down. this brings it more in line with antd components, and we could in the future change the onClose behaviour to do additional/different things.

src/js/components/Beacon/BeaconQueryUi.tsx
davidlougheed: could use the every() function of the array (plus a boolean cast of !! for formValues[field])

in this case, why is assembly ID not validated?

davidlougheed (changes requested):

src/js/components/Beacon/BeaconQueryUi.tsx
gsfk: The reasoning behind the variants form is that if you want to search variants you fill it in, and if you don't, you leave it blank. So the if the form is blank it's ignored, and none of the rules about required fields are applied. 

However, the form is almost never blank, since the assembly ID field is already filled for you in the (very common) case where there is only one assembly to search from. That's why it's treated separately from everything else.

src/js/components/Beacon/BeaconErrorMessage.tsx
davidlougheed: yes the state is fine, I mean that the prop on BeaconErrorMessage should be `onClose` and that should be bound to the state update function where this component is used

davidlougheed (approved):
lgtm

PR #132, opened 2023-12-06
perf: rm old files with dev build scripts + improve watch cpu usage

davidlougheed:


gsfk (approved):
Nice, webpack at idle is now ~3% instead of the previous ~70%

PR #133, opened 2023-12-11
fix inconsistent capitals in beacon helpext 

gsfk:
Fixes the inconsistent mix of "Enter..." and "enter" in Beacon variants form.


davidlougheed (approved):
lgtm

PR #144, opened 2024-01-11
beacon: use a different overview endpoint

gsfk:
Change beacon overview endpoint to `/overview` (Redmine [#1859](https://206.12.92.46/issues/1859))

### DIFF
--- a/src/js/features/beacon/beaconConfig.store.ts
+++ b/src/js/features/beacon/beaconConfig.store.ts
@@ -7,7 +7,7 @@ import { printAPIError } from '@/utils/error.util';
 export const getBeaconConfig = createAsyncThunk<BeaconConfigResponse, void, { state: RootState; rejectValue: string }>(
   'beaconConfig/getBeaconConfig',
   (_, { getState, rejectWithValue }) => {
-    const beaconInfoEndpoint = getState()?.config?.beaconUrl + '/overview';
+    const beaconInfoEndpoint = getState().config.beaconUrl + '/overview';
     return axios
       .get(beaconInfoEndpoint)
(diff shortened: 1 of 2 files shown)


src/js/features/beacon/beaconConfig.store.ts
```diff
@@ -7,7 +7,7 @@ import { printAPIError } from '@/utils/error.util';
 export const getBeaconConfig = createAsyncThunk<BeaconConfigResponse, void, { state: RootState; rejectValue: string }>(
   'beaconConfig/getBeaconConfig',
   (_, { getState, rejectWithValue }) => {
-    const beaconInfoEndpoint = getState()?.config?.beaconUrl + '/info';
+    const beaconInfoEndpoint = getState()?.config?.beaconUrl + '/overview';
```
davidlougheed, line 10: we should probably reject this if beaconUrl is undefined, but since the behaviour is already like this, this PR is good to merge

davidlougheed (approved):

src/js/features/beacon/beaconConfig.store.ts
gsfk: The UI isn't rendered until beaconUrl is defined, so it would be hard to get into this state.

davidlougheed: it's more that having the question marks sort of implies that it can somehow

also it's unneeded to have a `?` after `getState()` and `config` I think anyway, since getState.config should always return an object...

gsfk: This is probably a code copying crime on my part.

PR #147, opened 2024-02-14
patch condition for parsing beacon response

gsfk:
Patch condition for parsing the summary statistics part of the beacon response. Previously if `info` existed in the response, then so did `info.bento`, but this is no longer the case. See Redmine [#2080](https://206.12.92.46/issues/2080?issue_count=20&issue_position=1&next_issue_id=2068).


davidlougheed (approved):
lgtm

PR #156, opened 2024-03-19
Provenance/DATS fixes

gsfk:
- fix crashing on Provenance tab when some optional DATS sub-fields are missing
- fixes for multiple React warnings "Each child in a list should have a unique "key" prop"
- fixes for conditional rendering breaking ant design layout (remove `span`)


src/js/components/Provenance/Tables/ExtraPropertiesTable.tsx
```diff
@@ -20,17 +20,17 @@ const ExtraPropertiesTable = ({ extraProperties }: ExtraPropertiesTableProps) =>
           dataIndex: 'values',
           render: (_, { values }) =>
             values.map((v, i) => (
-              <>
-                <LinkIfUrl key={i} text={v.value.toString()} />
+              <div key={i}>
```
davidlougheed, line 23: this breaks how multiple links are rendered - instead, use `<React.Fragment key={i}>{...}</React.Fragment>` or similar

gsfk: For multiple entries, I get this, which doesn't look broken: 

Screenshot from 2024-03-21 09-17-41

davidlougheed: hmm I guess it doesn't render like I thought - i'm surprised the divs don't render as blocks but maybe there's some extra css at play. `<React.Fragment>` will still produce cleaner markup, since it will eliminate needless DOM elements.

gsfk: thanks, changed to `span`

davidlougheed: `<React.Fragment>` (or cleaner to `import { Fragment } from 'react'`) is still preferable, since it won't result in a `<span>` that isn't needed

gsfk: extra import vs a few spans? Anyway, changed to Fragment

src/js/components/Provenance/DatasetProvenance.tsx
```diff
@@ -38,31 +38,35 @@ const DatasetProvenance = ({ metadata, loading }: DatasetProvenanceProps) => {
 
         {/* --- CREATOR, PRIVACY, LICENSES, KEYWORD ---*/}
         {metadata.privacy || metadata.licenses.length || metadata.keywords.length ? (
-          <Descriptions style={{ paddingTop: '20px' }}>
+          <div style={{ display: 'flex', flexDirection: 'column', paddingTop: '20px' }}>
```
davidlougheed, line 41: is it possible to have this stay as one descriptions component like we talked about yesterday, or was there another reason that wouldn't work?

gsfk: only problem is forgetting to commit changes....

src/js/components/Provenance/Tables/DistributionsTable.tsx
```diff
@@ -58,7 +63,9 @@ const DistributionsTable = ({ distributions }: DistributionsTableProps) => {
     [td]
   );
 
-  return <BaseProvenanceTable dataSource={distributions} columns={columns} />;
+  return (
+    <BaseProvenanceTable dataSource={distributions} columns={columns} rowKey={(record) => record.access.landingPage} />
```
davidlougheed, line 67: in the types `landingPage` is required, but is it the case that it's occasionally missing in production instances?

gsfk: This was deliberate, since landing page is a required part of "access"

davidlougheed (approved):
lgtm

PR #160, opened 2024-03-25
style: fix top margins on search overview chart headings

davidlougheed:
also reduces minimum height to what is necessary.


gsfk (approved):
Looks good. Regarding `minHeight`: 
- there are no units (unlike maxWidth above it)
- is this still needed if it's simply the total of the contents? I get the same behaviour when removing it entirely.

PR #162, opened 2024-05-16
Site Sider

SanjeevLakhwani:
- Added a sider that replaces the tabbed dashboard in the previous version


src/js/components/SiteSider.tsx
```diff
@@ -0,0 +1,83 @@
+import React, { useState, useMemo, useCallback } from 'react';
+import { Layout, Menu } from 'antd';
+import type { MenuProps } from 'antd';
+import { PieChartOutlined, SearchOutlined, SolutionOutlined } from '@ant-design/icons';
+import { Flex } from 'antd';
+import Icon from '@ant-design/icons';
+import BeaconSvg from './Beacon/BeaconSvg';
+import { buildQueryParamsUrl } from '@/utils/search';
+import { useNavigate } from 'react-router-dom';
+import { useAppSelector, useTranslationDefault } from '@/hooks';
```
davidlougheed, line 10: please consolidate and order these inputs from `react-*` to `@ant-design*` to `@` imports to local imports

src/styles.css
```diff
@@ -17,3 +17,22 @@ body {
...
+}
+
+.logo {
+  height: 32px;
+  transition: transform 0.2s ease;
+}
+
+.logo-container.collapsed .logo {
+  transform: translateX(-70%);
+}
```
davidlougheed, line 38: missing trailing newline

davidlougheed (changes requested):
needs to account for custom logos

remove extra beacon file

gsfk (commented):
Side menu much prettier than the old tabs!

Perhaps a little late for this question, but what's the advantage of a "slider" over a simple side menu? I would think the main use case would be the ability to show a smaller menu when the sceen is smaller, but this pr doesn't do that.

gsfk (commented):
I'm not totally sold on the new header text... although the text appears in the same spot for each page, the page layout isn't consistent. Overview and Provenance are aligned with the component below them: 

Screenshot from 2024-05-23 10-15-24

... but Search & Beacon are not: 
gap

I also just find them ugly... I know that's not a particularly helpful comment. I wonder if this text wouldn't be better off in the header? Not sure.

header

src/js/components/SiteSider.tsx
```diff
@@ -0,0 +1,65 @@
...
+const BeaconLogo: React.FC<Partial<CustomIconComponentProps>> = (props) => <Icon component={BeaconSvg} {...props} />;
+
+const SiteSider: React.FC = () => {
+  const navigate = useNavigate();
+  const td = useTranslationDefault();
+  const [collapsed, setCollapsed] = useState(false);
+  const queryParams = useAppSelector((state) => state.query.queryParams);
+
+  const currentPage = getCurrentPage();
+  console.log({ currentPage });
```
davidlougheed, line 26: make this a proper debug log with information about whats being logged or remove it

src/js/hooks.ts
```diff
@@ -50,3 +50,16 @@ export const useBeaconWithAuthIfAllowed = () => {
     }
   }, [dispatch, hasPermission]);
 };
+
+// ################### ROUTER HOOKS ###################
+
+export const getCurrentPage = (): string => {
+  const location = window.location;
+  const pathArray = location.pathname.split('/');
+  const validPages = ['overview', 'search', 'beacon', 'provenance'];
```
davidlougheed, line 59: I don't like that this is hard-coded in multiple places...

src/js/hooks.ts
```diff
@@ -50,3 +50,16 @@ export const useBeaconWithAuthIfAllowed = () => {
     }
   }, [dispatch, hasPermission]);
 };
+
+// ################### ROUTER HOOKS ###################
+
+export const getCurrentPage = (): string => {
```
davidlougheed, line 56: technically not a hook - doesn't use `use` prefix and window.location/location/pathname don't trigger re-render

src/js/hooks.ts
```diff
@@ -50,3 +50,16 @@ export const useBeaconWithAuthIfAllowed = () => {
...
+  const location = window.location;
```
davidlougheed, line 57: weird re-assignment to do

src/styles.css
```diff
@@ -17,3 +17,22 @@ body {
...
   transform: scale(125%) !important;
 }
+
+.logo-container {
+  overflow: hidden;
+  height: 32px;
+  transition: width 0.2s ease;
+}
+
+.logo-container.collapsed {
```
davidlougheed, line 27: is this still needed?

davidlougheed (changes requested):
aesthetically, I think the bright white header text/button outlines are bit too harsh on the dark header background, but the visual hierarchy with the new sidebar being context for what page/project/dataset we are on works quite well.

maybe the contents of the sidebar should be sticky, so that we can change pages even if scrolled-down?

weird horizontal overflow/scroll, only on overview page on small screen (14" laptop)
Screenshot 2024-06-03 at 4 04 06 PM

gsfk (commented):
So....
- top of the screen is black but side is white?
- current tab indicated from menu only?

... I wasn't sure where we landed on either of these issues. 

The "Search" page does not behave consistently: if you load the main page and click Search it shows an overview, but it you load `/search` it shows zero results.

SanjeevLakhwani:
> * top of the screen is black but side is white?
This was done to correspond with the mock-up created previously .The hierarchy was changed from header now being the top-level component instead of the sidebar and the colors try to correspond with that. 
> * current tab indicated from menu only?
Yes as to reduce some clutter. The icons on the sidebar should now correctly represent what page one is on by being highlighted

davidlougheed (changes requested):
fix responsivity

davidlougheed (changes requested):

src/js/components/SiteHeader.tsx
```diff
@@ -32,44 +34,57 @@ const SiteHeader = () => {
...
+      <Flex align="center" justify="space-between">
+        <Space size="middle">
+          <a href="/">
+            <img
+              src="/public/assets/branding.png"
+              alt="logo"
+              style={{
+                height: '32px',
+                verticalAlign: 'middle',
+                marginLeft: '-15px',
```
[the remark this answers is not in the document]
SanjeevLakhwani: Header has a default y padding of 50px on both sides, I thought the logo looked better moved back a lil but on a second thought doesn't matter much so removed it now

davidlougheed (approved):
need to fix z-stacking / side-scrolling issues with narrow windows

image

davidlougheed (changes requested):
selected wrong opt

gsfk:
Looking good, I like the sticky menu. 

There's a conceptual change with chart width and the "manage charts" settings, the width setting is now really a max-width setting.

gsfk (commented):
Looking good, I like the sticky menu.

There's a conceptual change with chart width and the "manage charts" settings, the width setting is now really a max-width setting.

gsfk (approved):

davidlougheed (commented):
how will the new file for branding work with our existing instances that replace the logo?

SanjeevLakhwani:
> how will the new file for branding work with our existing instances that replace the logo?

I corrected the filenames to match the previous, so it should be the same as before

davidlougheed (changes requested):
icon on both left and right of portal button - on left, the old icon; on right, the link icon. otherwise looks good i think.

gsfk (commented):
The results pane jumps when switching between "Search" and "Beacon", it should be in the same place.

PR #163, opened 2024-05-23
Card description improvement

SanjeevLakhwani:
Card description now doesn't get cut off by the missing count. Done by reducing the rendering width for the description


gsfk (changes requested):
Not working for me. The `width: '300px'` setting is bigger than the size available when "missing" is present (only 262px for me). 

Also, and less important, the parameter `smallEllipses` should be spelled `smallEllipsis`

gsfk (commented):
Working now, but currently has a weird mix of spellings, including this odd typing: 

`smallEllipses: smallEllipsis`

gsfk (approved):

PR #164, opened 2024-06-03
Added histograms

SanjeevLakhwani:
- Now there is a distinction between bar charts and histogram supported by bento charts and katsu config (for labelling which is which)
- Bar charts are multi-colored while histograms are not
- There are new colors now

### DIFF
--- a/src/js/components/Overview/Chart.tsx
+++ b/src/js/components/Overview/Chart.tsx
@@ -58,7 +58,6 @@ const Chart = memo(({ chartConfig, data, units, id }: ChartProps) => {
           onClick={(d) => {
             navigate(`/${i18n.language}/search?${id}=${d.name}`);
           }}
-          colorTheme="new"
         />
       );
     case CHART_TYPE_CHOROPLETH: {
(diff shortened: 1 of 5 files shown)


src/js/components/Overview/Chart.tsx
```diff
@@ -30,6 +36,19 @@ const Chart = memo(({ chartConfig, data, units, id }: ChartProps) => {
...
       );
+    case CHART_TYPE_HISTOGRAM:
+      return (
+        <Histogram
+          units={units}
+          height={CHART_HEIGHT}
+          data={data}
+          preFilter={removeMissing}
+          dataMap={translateMap}
+          onClick={(d) => {
```
davidlougheed, line 47: factor out this function to a constant inside `Chart` since it's reused for bar/histogram

src/js/index.tsx
```diff
@@ -32,6 +31,8 @@ import { store } from './store';
 import { useBeaconWithAuthIfAllowed } from '@/hooks';
 
 import { PUBLIC_URL_NO_TRAILING_SLASH, CLIENT_ID, OPENID_CONFIG_URL, AUTH_CALLBACK_URL } from './config';
+import { NEW_BENTO_PUBLIC_THEME } from '@/constants/overviewConstants';
```
davidlougheed, line 34: inconsistent use of `@` vs `.`

davidlougheed (changes requested):

davidlougheed (approved):
lgtm

PR #166, opened 2024-06-13
New translations

SanjeevLakhwani:


src/js/components/Search/SearchResultsPane.tsx
```diff
@@ -19,7 +19,9 @@ const SearchResultsPane = ({
   experimentCount,
   experimentChartData,
 }: SearchResultsPaneProps) => {
-  const t = useTranslationDefault();
+  const td = useTranslationDefault();
+  const t = useTranslationCustom();
+  const translateMap = ({ x, y }: { x: string; y: number }) => ({ x: t(x), y });
```
davidlougheed, line 24: passed to a component, so wrap in a `useCallback`

src/public/locales/fr/translation_fr.json
```diff
@@ -12,5 +12,105 @@
...
+  "Experiments": "Expériences",
+  "Date of initial verbal consent (participant, legal representative or tutor), yyyy-mm-dd": "Date du consentement verbal initial (participant, représentant légal ou tuteur), aaaa-mm-jj",
+  "Covid severity": "Gravité de la Covid",
+  "Dead": "Décédé",
+  "Moderate": "Modéré",
+  "Uninfected": "Non infecté",
+  "Severe": "Sévère",
+  "Mild": "Léger",
+  "Mobility": "Mobilité",
+  "I have slight problems in walking about": "J’ai quelques problèmes à marcher",
```
davidlougheed, line 29: for the terms that come from BQC19, use the BQC19 translations instead of machine translation.

src/public/locales/en/translation_en.json
```diff
@@ -12,5 +12,105 @@
...
+  "Uninfected": "Uninfected",
+  "Severe": "Severe",
+  "Mild": "Mild",
+  "Mobility": "Mobility",
+  "I have slight problems in walking about": "I have slight problems in walking about",
+  "I have moderate problems in walking about": "I have moderate problems in walking about",
+  "I have severe problems in walking about": "I have severe problems in walking about",
+  "I am unable to walk about": "I am unable to walk about",
+  "Smoking Status": "Smoking Status",
+  "Smoking status": "Smoking status",
```
davidlougheed, line 34: why are both of these keys needed?

davidlougheed (changes requested):

src/public/locales/en/translation_en.json
SanjeevLakhwani: was a mistake removed it

davidlougheed (dismissed):
please make the translations have the same indent level

v-rocheleau (changes requested):
Looks good, only noticed an issue with the "in" translation.
By default we translate "in" => "dans", but this is grammatically incorrect in the Search context, where we have:

- "Age (in years)" => "Âge(dans years)"
   - should be "Âge (en années)"
- "Lab Test Results (in mg/L) => Résultats des tests de laboratoire (dans mg/L)"
   - should be "Résultats des tests de laboratoire (en mg/L)"

v-rocheleau (changes requested):
Close to being ok, there is still an issue with the "years" translation however:
image
This should be "(en années)"

v-rocheleau (approved):
Kept having the "en years" issue because I was using Bento's default translation file in `lib/public/translations`. After merging this we should update the default translations in the Bento repo.

PR #168, opened 2024-06-20
 Disable Click-to-Search for Non-Searchable Fields in Overview

SanjeevLakhwani:
This PR addresses the issue where clicking on non-searchable fields in the `bento_public` overview would trigger a search, leading to confusion as it returns results for the entire dataset.

There is now a clickable field on charts that is populated after the overview and query search fields are loaded and then there is a comparison. After this i done the charts are loaded to avoid unnecessary rerenders

### DIFF
--- a/src/js/components/BentoAppRouter.tsx
+++ b/src/js/components/BentoAppRouter.tsx
@@ -26,33 +26,19 @@ const BentoAppRouter = () => {
   const isAuthenticated = useIsAuthenticated();
 
   useEffect(() => {
-    const fetchInitialData = async () => {
-      try {
-        await Promise.all([dispatch(makeGetDataRequestThunk()), dispatch(makeGetSearchFields())]);
+    dispatch(makeGetConfigRequest()).then(() => dispatch(getBeaconConfig()));
+    dispatch(makeGetAboutRequest());
+    dispatch(makeGetDataRequestThunk());
+    dispatch(makeGetSearchFields()).then(() => dispatch(populateClickable()));
+    dispatch(makeGetProvenanceRequest());
+    dispatch(makeGetKatsuPublic());
+    dispatch(fetchKatsuData());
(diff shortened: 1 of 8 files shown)


src/js/components/BentoAppRouter.tsx
```diff
@@ -26,19 +26,31 @@ const BentoAppRouter = () => {
...
+
+        await Promise.all([dispatch(makeGetDataRequestThunk()), dispatch(makeGetSearchFields())]);
+
+        dispatch(populateClickable());
+
+        await dispatch(makeGetProvenanceRequest());
+        await dispatch(makeGetKatsuPublic());
+        await dispatch(fetchKatsuData());
+        await dispatch(fetchGohanData());
+        await dispatch(makeGetServiceInfoRequest());
```
gsfk, line 43: Lots of awaits, but are they all needed? The previous code only has a single action going through `then`, so why so many awaits now?

gsfk (commented):
No-click working fine. I'm puzzled by all the `awaits` though.

gsfk (commented):
Hmm.... the `followUpPromises` array actually launches all the dispatch calls rather than collecting them together. They have already run by the time you get to Promise.all.

Also, if the intention is to run these all concurrently note that `getBeaconConfig()` can be dispatched only _after_ `makeGetConfigRequest()`, that's why there's this line in the old code: 

> dispatch(makeGetConfigRequest()).then(() => dispatch(getBeaconConfig()));

If you run them concurrently you'll get race conditions and it will break randomly.

gsfk (commented):
Working okay.  

Failed intermittently for me on Chrome: sometimes produces only the arrow cursor and refuses to search anything. Happens pretty rarely so I can't give you a hint on how to reproduce it. 

The clickable area on some bar charts is pretty small, although perhaps this is outside the scope of this particular pr.

PR #171, opened 2024-07-29
feat: UI selector for project/dataset

SanjeevLakhwani:
Follows from #161
- improved over the existing redux over the metadata, reducing more stateful property
- Found a bug with `/callback` in index.js, it was never being triggered as a specific route as it was being accessed before by :lang param
- provided routing for project/dataset navigation by the format of /p/projecId/d/datasetId, rendered to the level needed
- modified the ui to select project/dataset 
- improvided on app structure by having separation of concerns with components and using Outlets and Loaders


src/js/components/ChooseProjectModal.tsx
```diff
@@ -0,0 +1,79 @@
...
+                <Typography.Text>{description}</Typography.Text>
+                <Typography.Title level={5}>Datasets</Typography.Title>
+                <List
+                  dataSource={datasets}
+                  bordered
+                  renderItem={(item) => (
+                    <Link to={`${baseURL}/p/${identifier}/d/${item.identifier}`}>
+                      <List.Item className="select-dataset-hover" key={item.identifier}>
+                        <List.Item.Meta
+                          avatar={<Avatar src="/public/assets/database.png" />}
```
davidlougheed, line 57: ideally this should be an existing icon / component, and at the very least an SVG

src/js/components/ChooseProjectModal.tsx
```diff
@@ -0,0 +1,79 @@
+import React, { useState } from 'react';
+import { Tabs, List, Avatar, Modal, Button, Space, Typography } from 'antd';
+import { useAppSelector } from '@/hooks';
+import { Link, useLocation } from 'react-router-dom';
+
+const ChooseProjectModal: React.FC<ChooseProjectModalProps> = ({ isModalOpen, setIsModalOpen }) => {
```
davidlougheed, line 6: React.FC is not ideal as a type: see https://github.com/facebook/create-react-app/pull/8177

src/js/components/ChooseProjectModal.tsx
```diff
@@ -0,0 +1,79 @@
...
+            <Button>Select Default View</Button>
+          </Link>
+        </Space>
+      }
+    >
+      <Tabs
+        tabPosition="left"
+        activeKey={selectedProject}
+        onChange={(key) => setSelectedProject(key)}
+        items={projects.map(({ identifier, title, datasets, description }) => {
```
davidlougheed, line 36: I'm not a big fan of the way these are rendered - the visual hierarchy is unclear, as is the "cancel" button's purpose and the "select default view" button.

davidlougheed (changes requested):
please fix merge conflicts.

src/js/components/ChooseProjectModal.tsx
SanjeevLakhwani: I am not sure why it needs to be an svg. Given the rendering size of the icon, a png works fine.

src/js/components/ChooseProjectModal.tsx
SanjeevLakhwani: How about this: 
image

gsfk: As a sometimes confused user, I'm still a bit puzzled at the user actions needed to choose between viewing a whole project and just viewing a particular dataset. Intuitively you'd think they would be more similar.

src/styles.css
```diff
@@ -22,3 +22,12 @@ body {
 .ant-layout-sider-trigger {
   border-right: 1px solid #f0f0f0;
 }
+
+.select-project-title:hover {
+  color: white !important;
```
davidlougheed, line 27: add a transition for the color change

src/styles.css
```diff
@@ -22,3 +22,12 @@ body {
...
+  cursor: pointer;
+}
+
+.select-dataset-hover:hover {
+  background-color: aliceblue !important;
```
davidlougheed, line 32: add a transition for the color change (~0.15s)

src/styles.css
davidlougheed: (maybe ~0.15s)

davidlougheed (changes requested):
image

image


still many rendering glitches here:

* unneeded margin above project title
* project title and datasets subheading should not be the same heading level
* button margin for "clear selection"
* hover background for dataset goes outside of border radius bounds
* text in header isn't aligned with icon vertically
* I think the node heading should be a bit bigger, and then so should this selector heading in the header

davidlougheed:
image
i also get this when I load – no routes match `/`

davidlougheed:
image

src/js/components/ChooseProjectModal.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { useAppSelector, useTranslationCustom, useTranslationDefault } from '@/hooks';
+import { Link, useLocation } from 'react-router-dom';
+import { FaDatabase } from 'react-icons/fa';
+
+const ChooseProjectModal = ({ isModalOpen, setIsModalOpen }: ChooseProjectModalProps) => {
+  const td = useTranslationDefault();
+  const t = useTranslationCustom();
+  const location = useLocation();
+  const { projects, selectedScope } = useAppSelector((state) => state.metadata);
+  const [selectedProject, setSelectedProject] = useState(selectedScope.project ?? projects[0].identifier);
```
davidlougheed, line 12: breaks with no projects (projects[0] is undefined)

src/js/components/SiteHeader.tsx
```diff
@@ -40,16 +58,28 @@ const SiteHeader = () => {
...
+            level={5}
+            style={{ fontSize: '16px', margin: 0, lineHeight: '64px', color: 'lightgray' }}
+            onClick={() => setIsModalOpen(true)}
+          >
+            <ProfileOutlined style={{ marginRight: '5px', fontSize: '16px' }} />
+
+            {selectedScope.project && scopeProps.projectTitle}
+            {scopeProps.datasetTitle ? ` / ${scopeProps.datasetTitle}` : ''}
+          </Typography.Title>
+          )
```
davidlougheed, line 81: stray ')'

davidlougheed (changes requested):

src/js/types/metadata.ts
```diff
@@ -0,0 +1,36 @@
+interface DiscoveryOverview {}
+interface DiscoverySearch {}
+interface DiscoveryFields {}
+interface DiscoveryRules {}
```
davidlougheed, line 4: what's the plan for these types? I though there were already types for this stuff?

src/js/types/metadata.ts
```diff
@@ -0,0 +1,35 @@
...
+  description: string;
+  discovery: Discovery;
+  datasets: Dataset[];
+}
+
+export interface Dataset {
+  identifier: string;
+  title: string;
+  description: string;
+  discovery: Discovery;
```
davidlougheed, line 24: these can be null, no?

src/js/types/metadata.ts
```diff
@@ -0,0 +1,35 @@
...
+  search: DiscoverySearch[];
+  fields: DiscoveryFields;
+  rules: DiscoveryRules;
+}
+
+export interface Project {
+  identifier: string;
+  title: string;
+  description: string;
+  discovery: Discovery;
```
davidlougheed, line 16: these can be null, no?

davidlougheed (changes requested):

davidlougheed (approved):
lgtm

PR #176, opened 2024-08-26
lint: enforce import type

SanjeevLakhwani:


davidlougheed (commented):
would prefer fixStyle: inline-type-imports - should change in bento_web as well.

davidlougheed:
https://typescript-eslint.io/rules/consistent-type-imports/

src/js/components/Beacon/VariantsForm.tsx
```diff
@@ -1,9 +1,10 @@
-import React, { CSSProperties } from 'react';
+import type { CSSProperties } from 'react';
+import React from 'react';
```
davidlougheed, line 2: remove, not needed anyway

src/js/hooks.ts
```diff
@@ -1,11 +1,14 @@
-import React, { useCallback, useEffect, useState } from 'react';
-import { TypedUseSelectorHook, useDispatch, useSelector } from 'react-redux';
+import type React from 'react';
```
davidlougheed, line 1: remove, not needed anyway

davidlougheed (changes requested):

src/js/hooks.ts
davidlougheed: (convert `React.*` to imports)

src/js/components/Beacon/VariantsForm.tsx
davidlougheed: (convert `React.*` to imports)

davidlougheed (approved):
lgtm

PR #182, opened 2024-09-12
provided a loader on overview when changing scope

SanjeevLakhwani:


davidlougheed:
there's still a lag between loading and when the charts actually re-render - i think maybe because chart data is copied/modified vs what is fetched and reconciled with localstorage etc?

davidlougheed:
it doesn't happen all the time, but it does mean there's something not ideal about the way this works

davidlougheed:
however, it's an improvement – so i'll approve and we'll figure out intricacies later.

davidlougheed (approved):

PR #185, opened 2024-09-18
chore: set up new JSX transform and remove "import React"

SanjeevLakhwani:
Set up new JSX transform and remove "import React" from files

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -19,6 +19,7 @@ module.exports = {
     'plugin:@typescript-eslint/recommended',
     'plugin:prettier/recommended',
     'plugin:react-hooks/recommended',
+    'plugin:react/jsx-runtime',
   ],
   rules: {
     'react/prop-types': 'off', // disable prop-types since we're using TypeScript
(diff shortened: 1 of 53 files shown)


.eslintrc.js
```diff
@@ -38,6 +39,7 @@ module.exports = {
       },
     ],
     semi: ['error', 'always'],
+    'react/react-in-jsx-scope': 'off',
```
v-rocheleau, line 42: Looking at the doc for jsx-runtime, Im confused as to why `"react/jsx-uses-react": "off"` is not required here too, any idea?

davidlougheed:
needs build fix

.eslintrc.js
SanjeevLakhwani: no clue

src/js/components/Overview/Chart.tsx
```diff
@@ -1,4 +1,5 @@
-import React, { memo } from 'react';
+// import React, { memo } from 'react';
```
davidlougheed, line 1: rm?

davidlougheed (changes requested):

davidlougheed (approved):

PR #186, opened 2024-09-18
removed example.server.env and its references

SanjeevLakhwani:


gsfk (commented):
you can remove the cruft from gitignore and dockerignore also

gsfk (approved):

PR #188, opened 2024-09-23
forced boolean type for beacon_ui_enabled

SanjeevLakhwani:


gsfk (changes requested):
Beacon on/off works as expected but I find this fix confusing.

For me the expected fix was to keep BEACON_UI_ENABLED as a boolean everywhere, and wrap _only_  process.env.BENTO_BEACON_UI_ENABLED in stringToBoolean(), ie:

```
BEACON_UI_ENABLED = BENTO_PUBLIC_CONFIG.BEACON_UI_ENABLED ?? stringToBoolean(process.env.BENTO_BEACON_UI_ENABLED);

```
 

.. ie you cast to boolean when you first see the value, instead of passing the string along. Otherwise you get a confusing situation where

Config.BEACON_UI_ENABLED is a string
BEACON_UI_ENABLED is a boolean

Note that there is another boolean in config (`TRANSLATED`) that probably has similar issues. 

(also, possibly beyond the scope of this pr, why are there two values with the same name? Why do we export a separate BEACON_UI_ENABLED variable instead of simply exporting the config object?)

Another issue that's probably not specific to this pr is the expected behaviour when asking for routes that don't exist (like asking for `/beacon` on an instance with beacon turned off): previously if you asked for a non-existent route, you would be returned to the default page (Overview), now you just get a white screen... not a crash, just ... nothing...

gsfk (approved):

davidlougheed (changes requested):
create_config_prod is made to be ran separate from typescript, and so it should not include any code from `src/` - keep a separate definition inside `create_config_prod.js`

davidlougheed (changes requested):
make `parseBoolean` in `create_config_prod` check the same values as stringToBoolean in utils/strings

davidlougheed (approved):

PR #189, opened 2024-09-25
chore: rm proptypes for DownloadDats component

davidlougheed:
this is more narrowly specified by the TypeScript prop type anyway.


gsfk (approved):
DATS download broken in main, so not really possible to test

gsfk:
.. was supposed to be comment rather than approval..

PR #190, opened 2024-09-25
chore: remove unused code for iframed Bento Public

davidlougheed:
I think this was some experiment for use in a BQC19 website iframe attempt, but to my knowledge this isn't used anywhere (confirmed with David B).


gsfk (approved):
I guess it used to live here? https://www.quebeccovidbiobank.ca/bento

PR #191, opened 2024-09-27
feat(search): show individual match results when available

davidlougheed:


gsfk (commented):
Works as expected. I'm assuming more stuff will be filled in later. 

I had to click "Sign in" on the portal page that popped up, even though I was already signed in, although I wasn't forced to login a second time. But I assume this too won't be an issue later.

davidlougheed:
yeah the idea is to have something relatively straightforward to show that we're moving towards integration of auth and gradual permissions, and to slowly merge it with beacon/the explore page as we go.

src/styles.css
v-rocheleau, line 1: It was not clear to me at first that the `Individuals` element was clickable, I think a hover class would help with that.

v-rocheleau (changes requested):

src/js/components/Search/SearchResults.tsx
```diff
@@ -8,6 +8,7 @@ const SearchResults = () => {
   const experimentCount = useAppSelector((state) => state.query.experimentCount);
   const experimentChartData = useAppSelector((state) => state.query.experimentChartData);
   const individualCount = useAppSelector((state) => state.query.individualCount);
+  const individualMatches = useAppSelector((state) => state.query.individualMatches);
```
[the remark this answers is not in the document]
davidlougheed: i agree

src/styles.css
davidlougheed: sounds good!

v-rocheleau (approved):
LGTM!

PR #192, opened 2024-09-30
fix: DATS download + as attachment 

davidlougheed:
Also removes the prop types for the DownloadDats component.


davidlougheed:
@gsfk goes with https://github.com/bento-platform/katsu/pull/545

gsfk (approved):

PR #193, opened 2024-09-30
refact(provenance): filter by scope and deduplicate redux store

davidlougheed:
Removes the provenance store, since all of that data is already in the projects response from Katsu.


v-rocheleau (approved):
lgtm!

PR #199, opened 2024-10-09
style: tweak styling for about HTML content

davidlougheed:


src/styles.css
```diff
@@ -60,15 +60,35 @@ body {
   color: #91caff;
 }
 
+
+/* BEGIN about HTML styling */
+
+/*
+Header elements which are the first child of the about content container shouldn't have the default Ant Design top
```
SanjeevLakhwani, line 67: The about html does not inherit antd styles for what I know so this shouldnt be an issue.

src/styles.css
```diff
@@ -60,15 +60,35 @@ body {
...
+margin
+ */
+.about-content h2:first-child, h3:first-child {
```
SanjeevLakhwani, line 70: why just h2 and h3, why not h1?

src/styles.css
davidlougheed: it does seem to on my end and is visible for example with RENATA: https://renata.bento.sd4h.ca/en/

src/styles.css
davidlougheed: was just doing what is used but you're right, should be all

src/styles.css
```diff
@@ -60,15 +60,32 @@ body {
   color: #91caff;
 }
 
+
+/* BEGIN about HTML styling */
+
+/* Header elements which are the first child of the about content container shouldn't have the default top margin */
+.about-content h1:first-child, h2:first-child, h3:first-child, h4:first-child, h5:first-child, h6:first-child {
```
SanjeevLakhwani, line 67: We can also consider maybe just affecting the first child of about content instead of just the headers as first child?

SanjeevLakhwani (commented):
lgtm, just the comment

src/styles.css
davidlougheed: done

PR #201, opened 2024-10-09
style: beacon network styling

davidlougheed:
* Non-fixed-width network nodes
* Use Ant column/row system for nodes
* Removes some unnecessary CSS
* Fixes up titled search results display
* Adds logo alt text


gsfk (commented):
Nice, thanks!

Only weird behaviour I could produce is when screen width is very narrow and one beacon name is long. But this puts us into telephone screen dimensions where there are existing issues already. 



Screenshot from 2024-10-10 13-25-46

gsfk (approved):

PR #202, opened 2024-10-10
refact: useBeaconNetwork hook + some translation

davidlougheed:


src/js/features/beacon/networkQuery.store.ts
```diff
@@ -29,7 +29,7 @@ const queryBeaconNetworkNode = createAsyncThunk<
   BeaconQueryResponse,
   QueryToNetworkBeacon,
   { state: RootState; rejectValue: string }
->('queryBeaconNetworkNode', async ({ _beaconId, url, payload }, { rejectWithValue }) => {
+>('queryBeaconNetworkNode', async ({ url, payload }, { rejectWithValue }) => {
```
gsfk, line 32: I went back and forth on including this a parameter or just handling it elsewhere, but removing it entirely didn't occur to me, it's wild that this still works.

gsfk (approved):
Thanks, seems good.

src/js/features/beacon/networkQuery.store.ts
davidlougheed: since it's a destructured object as a param, you only need to destructure the properties of the object that you need

PR #205, opened 2024-10-21
feat: help text for Bento concepts in counts display

davidlougheed:
also refactors translations for entities to be more structured.


gsfk (approved):
Seems to work fine. 

I was a bit confused by how the translation structure is done with string literals, especially where it's explicit in `title` but then implicit in `help`, but that's probably on me. 

There will be a few merge conflicts with the Beacon Network UI stuff, although nothing too crazy.

PR #206, opened 2024-10-21
refact: combine beacon and network double-slices

davidlougheed:


gsfk (approved):
Nice... seems to work well, thanks for the hook!

Only thing I thought was overkill was the extra boilerplate in `errorMsgOrDefault` but I see I already let that slip by.

PR #211, opened 2024-10-25
refact: single translation function with fallback + missing i18n

davidlougheed:


gsfk (approved):
Seems good. 

I had to look closely at the translations, and that brought up lots of issues, but I don't think any of them are specific to this branch. I'll write a ticket for that instead.

SanjeevLakhwani (approved):
lgtm

PR #212, opened 2024-11-06
Added New Chart Manager Drawer

SanjeevLakhwani:
This makes the button's meaning more explicitly visualised.


davidlougheed (approved):
lgtm

PR #221, opened 2024-11-13
fix: chart clicking in French

SanjeevLakhwani:
fixes: https://redmine.c3g-app.sd4h.ca/issues/2379
issue: On translated bars/slice on chart on click search wasn't working correctly as it used to redirect to the french label for the search but the backend only supports english keys, now the requests are translation agnostic.


davidlougheed (approved):
lgtm

gsfk (approved):

PR #223, opened 2024-11-13
refact: use i18n for data type / entity labels

davidlougheed:


src/js/components/Beacon/BeaconCommon/BeaconQueryFormUi.tsx
```diff
@@ -233,7 +233,7 @@ const BeaconQueryFormUi = ({
             {hasVariants && (
               <Col xs={24} lg={12}>
                 <Card
-                  title={t('entities.Variants')}
+                  title={t('entities.variant', { count: 100 })} // false count – just need the highest form of plural
```
v-rocheleau, line 236: Nice didn't know about the plural feature.
Since these objects are passed to the `t` function a bunch all across the app, I think we could use constants instead to make it easier to write and understand. Something like:
```typescript
// in a constants file
export const T_PLURAL_COUNT = { count: 100 };
export const T_SINGULAR_COUNT = { count: 1 };

// usage
title={t('entities.variant', T_PLURAL_COUNT)}
```
I think this way it would be clear even without the comment.

src/js/features/dataTypes/dataTypes.store.ts
```diff
@@ -1,34 +1,45 @@
...
+      if (!cond) console.debug(`makeGetDataTypes() was attempted, but a prior attempt gave status: ${status}`);
+      return cond;
+    },
+  }
+);
 
 export type DataTypesState = {
-  isFetching: boolean;
-  dataTypes: object;
+  status: RequestStatus;
```
v-rocheleau, line 36: Nice, I like this better than a simple boolean, gives more information.

src/public/locales/en/default_translation_en.json
```diff
@@ -6,16 +6,21 @@
   "Portal": "Portal",
   "missing": "missing",
   "entities": {
-    "Phenopackets": "Clinical Data",
+    "phenopacket_one": "Phenopacket",
+    "phenopacket_other": "Clinical Data",
+    "individual_one": "Individual",
+    "individual_other": "Individuals",
     "Individual": "Individual",
-    "Individuals": "Individuals",
```
v-rocheleau, line 11: Why not `individual_one` and `individual_other` like the other entities? Mismatches with the FR translation

v-rocheleau (changes requested):
Small requests, otherwise looks good

src/public/locales/en/default_translation_en.json
davidlougheed: oops - forgot

src/js/components/Beacon/BeaconCommon/BeaconQueryFormUi.tsx
davidlougheed: good idea!

v-rocheleau (approved):
Lgtm!

PR #225, opened 2024-11-21
Dataset Catalog

SanjeevLakhwani:
Added a dataset catalog which is displayed in the overview tab when no project is selected


src/js/components/Overview/PublicOverview.tsx
```diff
@@ -59,6 +60,8 @@ const PublicOverview = () => {
     saveToLocalStorage(sections);
   }, [sections]);
 
+  if (!selectedProject) return <Catalogue />;
```
davidlougheed, line 63: does this work with single-project nodes? I know there's an auto-select, but does the catalog flicker in and out when the node only has one project? should it be loader-aware since below this early return skips the loading...

src/js/components/Provenance/Catalogue/Catalogue.tsx
```diff
@@ -0,0 +1,16 @@
+import { Space } from 'antd';
+import CatalogueCard from '@/components/Provenance/Catalogue/CatalogueCard';
+import { useMetadata } from '@/features/metadata/hooks';
+
+const Catalogue = () => {
+  const { projects } = useMetadata();
+  return (
+    <Space direction="vertical" style={{ width: '100%' }}>
+      {projects.map((project) => (
+        <CatalogueCard project={project} key={project.title} />
```
davidlougheed, line 10: key should be identifier.

src/js/components/Provenance/Catalogue/CatalogueCard.tsx
```diff
@@ -0,0 +1,98 @@
...
+        <Paragraph ellipsis={{ rows: 1, tooltip: { title: ISODateToString(project.updated) } }}>
+          {ISODateToString(project.updated)}
+        </Paragraph>
+      ),
+      span: 1.5,
+    },
+  ];
+
+  return (
+    <Card key={project.title} style={{ boxShadow: '0 2px 10px rgba(0,0,0,0.05)', height: '260px' }}>
```
davidlougheed, line 58: keyword should be identifier.

src/js/components/Provenance/Catalogue/CatalogueCard.tsx
```diff
@@ -0,0 +1,98 @@
...
+              {extraKeywordCount !== 0 && <Typography.Text>+{extraKeywordCount} more</Typography.Text>}
+            </div>
+            <Descriptions items={projectInfo} />
+          </Space>
+        </div>
+        <div style={{ width: '50%', maxWidth: '600px' }}>
+          <Typography.Title level={4} style={{ marginTop: 0 }}>
+            Datasets
+          </Typography.Title>
+          <Carousel arrows autoplay style={{ border: '1px solid lightgray', borderRadius: '7px' }}>
```
davidlougheed, line 88: i don't like autoplay here - it will get distracting with large # of projects/datasets and a bit frustrating to use.

src/js/utils/strings.ts
```diff
@@ -5,3 +5,10 @@ export const stringToBoolean = (s: string | undefined) =>
 
 export const stringIsDOI = (s: string) => !!s.match(DOI_PATTERN);
 export const stringIsURL = (s: string) => !!s.match(URL_PATTERN);
+
+export const ISODateToString = (d: string) =>
```
davidlougheed, line 9: use lowercase `iso` for starting a function name

src/styles.css
```diff
@@ -119,6 +119,14 @@ body {
   gap: 20px;
 }
 
+.slick-arrow {
+  color: black !important; /* override slick-next/slick-prev color */
+}
+
+.slick-dots li button{
```
davidlougheed, line 126: space between button and `{`

src/styles.css
```diff
@@ -119,6 +119,14 @@ body {
   gap: 20px;
 }
 
+.slick-arrow {
+  color: black !important; /* override slick-next/slick-prev color */
```
davidlougheed, line 123: does it need to be !important?

src/styles.css
```diff
@@ -138,3 +146,4 @@ body {
     gap: 8px;
   }
 }
+
```
davidlougheed, line 149: needed or no

src/js/components/Provenance/Catalogue/CatalogueCard.tsx
```diff
@@ -0,0 +1,98 @@
...
+    selectedKeywords.push(keyword);
+    totalCharacters += keyword.length;
+  }
+
+  const extraKeywordCount = Math.max(keywords.length - selectedKeywords.length, 0);
+
+  const projectInfo = [
+    {
+      key: '1',
+      label: 'Created',
```
davidlougheed, line 32: translation

src/js/components/Provenance/Catalogue/CatalogueCard.tsx
```diff
@@ -0,0 +1,98 @@
...
+          }}
+        >
+          {ISODateToString(project.created)}
+        </Paragraph>
+      ),
+      span: 1.5,
+    },
+    {
+      key: '2',
+      label: 'Updated',
```
davidlougheed, line 47: translation

src/js/components/Provenance/Catalogue/CatalogueCard.tsx
```diff
@@ -0,0 +1,98 @@
...
+                </Tag>
+              ))}
+              {extraKeywordCount !== 0 && <Typography.Text>+{extraKeywordCount} more</Typography.Text>}
+            </div>
+            <Descriptions items={projectInfo} />
+          </Space>
+        </div>
+        <div style={{ width: '50%', maxWidth: '600px' }}>
+          <Typography.Title level={4} style={{ marginTop: 0 }}>
+            Datasets
```
davidlougheed, line 86: translation

src/js/components/Overview/PublicOverview.tsx
```diff
@@ -18,6 +18,7 @@ import { useAppSelector } from '@/hooks';
 import { useSelectedProject, useSelectedScope } from '@/features/metadata/hooks';
 import { useTranslation } from 'react-i18next';
 import { RequestStatus } from '@/types/requests';
+import Catalogue from '@/components/Provenance/Catalogue/Catalogue';
```
davidlougheed, line 21: move with other component imports

davidlougheed (changes requested):

src/styles.css
SanjeevLakhwani: I would say needed, but on my end there are two newlines so going to edit that.

gsfk (commented):
Carousel autoplay is a bit.... I don't know, aggresive? I think we typically avoid animations, Also: https://shouldiuseacarousel.com/

For projects with a single dataset, I still get left and right arrows, is this intentional?

src/js/components/Provenance/Catalogue/CatalogueCard.tsx
```diff
@@ -0,0 +1,101 @@
...
+        </Paragraph>
+      ),
+      span: 1.5,
+    },
+  ];
+
+  return (
+    <Card key={project.identifier} style={{ boxShadow: '0 2px 10px rgba(0,0,0,0.05)', height: '260px' }}>
+      <Flex justify="space-between">
+        <div style={{ width: '50%', paddingRight: '10px' }}>
```
davidlougheed, line 63: the hard-coded fixed percent width isn't responsive.

src/js/components/Provenance/Catalogue/CatalogueCard.tsx
```diff
@@ -0,0 +1,101 @@
...
+        <Paragraph ellipsis={{ rows: 1, tooltip: { title: isoDateToString(project.updated) } }}>
+          {isoDateToString(project.updated)}
+        </Paragraph>
+      ),
+      span: 1.5,
+    },
+  ];
+
+  return (
+    <Card key={project.identifier} style={{ boxShadow: '0 2px 10px rgba(0,0,0,0.05)', height: '260px' }}>
```
davidlougheed, line 61: re-use standard box shadow?

src/js/components/Provenance/Catalogue/CatalogueCarouselDataset.tsx
```diff
@@ -0,0 +1,19 @@
+import { Card, Typography } from 'antd';
+import type { Dataset } from '@/types/metadata';
+import { useTranslationFn } from '@/hooks';
+
+const CatalogueCarouselDataset = ({ dataset }: { dataset: Dataset }) => {
+  const t = useTranslationFn();
+  return (
+    <Card style={{ height: '180px' }}>
```
davidlougheed, line 8: instead of being fixed height, this should expand to fill whatever container it is in

src/js/components/Provenance/Catalogue/Catalogue.tsx
```diff
@@ -0,0 +1,16 @@
+import { Space } from 'antd';
+import CatalogueCard from '@/components/Provenance/Catalogue/CatalogueCard';
+import { useMetadata } from '@/features/metadata/hooks';
+
+const Catalogue = () => {
+  const { projects } = useMetadata();
+  return (
+    <Space direction="vertical" style={{ width: '100%' }}>
```
davidlougheed, line 8: the gap here isn't big enough visually between `CatalogueCard`s - should be the same as cards in the overview, and scale the same with screen width (like with the overview)

src/js/components/Provenance/Catalogue/CatalogueCard.tsx
```diff
@@ -0,0 +1,101 @@
...
+      <Flex justify="space-between">
+        <div style={{ width: '50%', paddingRight: '10px' }}>
+          <Space direction="vertical">
+            <Typography.Title level={5} style={{ marginTop: 0 }}>
```
davidlougheed, line 65: this title level is below the one used for datasets, which is not correct hierarchically .

davidlougheed (changes requested):

SanjeevLakhwani (commented):
The language switcher is broken on the catalog. Should get resolved soon.

src/js/components/Provenance/Catalogue/CatalogueCard.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { isoDateToString } from '@/utils/strings';
+import { useTranslationFn } from '@/hooks';
+import { BOX_SHADOW } from '@/constants/overviewConstants';
+import Dataset from '@/components/Provenance/Catalogue/Dataset';
+
+const { Paragraph, Text, Title } = Typography;
+
+const MAX_KEYWORD_CHARACTERS = 50;
+
+const CatalogueCard = ({ project }: { project: Project }) => {
```
davidlougheed, line 14: I think for consistency this should be called Project. we can then later split it into multiple view modes, like with Dataset.

src/js/utils/strings.ts
```diff
@@ -1,7 +1,17 @@
...
 
 export const stringToBoolean = (s: string | undefined) =>
   ['true', 't', '1', 'yes'].includes((s || '').toLocaleLowerCase());
 
 export const stringIsDOI = (s: string) => !!s.match(DOI_PATTERN);
 export const stringIsURL = (s: string) => !!s.match(URL_PATTERN);
+
+export const isoDateToString = (d: string, lang?: string) => {
+  const dateLang = lang === 'fr' ? 'fr-CA' : 'en-US';
+  return new Date(d).toLocaleString(dateLang, {
```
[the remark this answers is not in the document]
SanjeevLakhwani: I gets accessed before i18n is initialized causing issues.

src/js/components/Provenance/Catalogue/CatalogueCard.tsx
SanjeevLakhwani: I would rather change, Dataset to CatalogDataset as Dataset/Project by itself can mean multiple things in public context by conflicting with overview. I do understand that it is in a folder but the naming does reduce the readability.

davidlougheed: catalogdataset to me doesn't make sense, as the dataset isn't just used in the catalog

src/js/utils/strings.ts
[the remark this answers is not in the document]
davidlougheed: but can it be passed into the function elsewhere, thus mitigating the "if french else assume english" extra logic?

src/js/components/Provenance/Catalogue/CatalogueCard.tsx
SanjeevLakhwani: I mean it in the form that a "Dataset" could mean multiple things while at least prefixing catalog tells the purpose. If you think otherwise still, lmk, I shall change it.

src/js/utils/strings.ts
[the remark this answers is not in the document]
SanjeevLakhwani: I can create a utility function for it to separate logic., but I think such logic also exists on the initial render of the website (except there is an extra log statement for saying unsupported language).

davidlougheed (commented):
looks good to me, but I'd like @gsfk to take a look before this gets merged.

gsfk:
Looks good. 

I had some permissions issues (unrelated to the pr it seems) and noticed that public tends to make scope errors in cases where the user has insufficient permissions. For example, in this pr: 
- load the front page (the catalogue)
- choose Catalogue > Some Project > Some Dataset > Overview without good permissions you get the overview for the entire instance. The scope url is correct, and reloading get you a correct view, looks like data is just stale. 

However, this happens on current main as well, so doesn't necessarily need to be fixed in this branch.

PR #227, opened 2024-11-21
Added onclick pointer to bar chart/histogram

SanjeevLakhwani:
This was removed when onClick started to work with on click on the possible render area for a bar as it caused multiple complications. Now it added back but only for bars even though you are allowed to click around it also.


gsfk (commented):
Bit of a conflict between actual clickable area (entire height of bar area) and area where you get the pointing hand cursor (filled-in portion of the bar). But I'm guessing this is a hassle to fix?

davidlougheed:
There's a way to add a cursor: pointer to the whole thing, but that ends up overriding hovers without click too. Not sure if there's a simple solution but this does help a bit.

davidlougheed (approved):
lgtm

PR #229, opened 2024-11-22
feat: UI query:data permissions awareness for counts

davidlougheed:


davidlougheed:
I can change it back, I hated the non-em dashes but that's personal preference

For beacon, I think we can maybe determine this from permissions + the rules count threshold? instead of trying to ascertain it from response

gsfk:
Ah, no point putting dashes back if you hate them.

PR #233, opened 2024-11-26
chore: address confusing UX for network chromosome/assembly

davidlougheed:
now chromosome won't be disabled/"select an assembly"'d for beacon network, since there isn't a contig dropdown


gsfk (commented):
given the layout of the form, it's probably expected I fill in "chromosome" first and "assembly ID" second, but when I choose an assembly it blanks my chromosome choice.....

davidlougheed:
forgot about that, thanks for the catch.

gsfk (approved):

PR #235, opened 2024-11-27
refact(overview): use hook for calculating chart clickability

davidlougheed:


src/js/components/Overview/PublicOverview.tsx
```diff
@@ -94,7 +94,7 @@ const PublicOverview = () => {
         <Col flex={1}>
           {displayedSections.map(({ sectionTitle, charts }, i) => (
             <div key={i} className="overview">
-              <OverviewSection title={sectionTitle} chartData={charts} />
+              <OverviewSection title={sectionTitle} chartData={charts} searchableFields={searchableFields} />
```
SanjeevLakhwani, line 97: Feels like an unnecessary prop drill. I would suggest searchableFields to be an object with keys as sections and values as the searchable fields in that section. I do understand that calling it here reduces to calls to useSearchableFields but also we are passing irrelevant data to sections.

SanjeevLakhwani (changes requested):

src/js/components/Overview/PublicOverview.tsx
davidlougheed: i agree the prop drilling is a bit much. was avoiding multiple hook calls but i doubt the performance hit is that bad.

davidlougheed: actually, since query sections != overview sections, we do need to pass everything

SanjeevLakhwani (approved):
lgtm.

PR #238, opened 2024-12-11
fix: selector indicator for dataset scope selection

davidlougheed:
by fixing, this means that datasets are now highlighted on selection in the scoping modal.

also refactors validProjectDataset to use a cleaner lookup into an object instead of an array.


gsfk (commented):
Works well. Do we need both `projects` and `projectsByID`?

PR #241, opened 2024-12-17
add scoping to beacon UI

gsfk:
First pass at scope-aware Beacon UI.

Requires beacon [scoping PR](https://github.com/bento-platform/bento_beacon/pull/120).


davidlougheed (commented):
looks good to me (pending Beacon as mentioned)

gsfk:
> looks good to me (pending Beacon as mentioned)

Still has a few too many renders... it loads full scope first, then re-renders with selected scope.

davidlougheed:
there's a `scopeSet` property that may help here (on the selectedScope return object)

src/js/components/Beacon/BeaconCommon/BeaconQueryFormUi.tsx
```diff
@@ -67,12 +70,18 @@ const BeaconQueryFormUi = ({
...
+      };
+      if (scope.dataset) {
+        payload['query']['datasets'] = { datasetIds: [scope.dataset] };
+      }
+      return payload;
+    },
+    [scope]
   );
 
   const launchEmptyQuery = useCallback(
```
davidlougheed, line 87: can't comment on the below, but the empty query should only launch (with the useeffect below) when the scope has been successfully set from the URL via `scopeSet`. I think this should fix some of the re-rendering.

davidlougheed (approved):
lgtm

PR #242, opened 2025-01-13
feat: URL context-aware sider, header, router

davidlougheed:


gsfk (commented):
Working well, makes the connection between scope and menu items a bit more clear.

Would be nice eventually to have the connection more clear visually (the scope and menu items are far apart from each other on the page)... but I don't have a concrete suggestion. Maybe the data catalogue will clear this up.

davidlougheed:
I think it'll be a multi-PR process with the catalogue and perhaps rearranging the scope menu / replacing it with something else.

PR #243, opened 2025-01-17
refact: dataset catalogue styling + responsiveness

davidlougheed:
this essentially addresses what I was going to put in most of the review. still a work in progress, but let me know what you think @SanjeevLakhwani.


src/js/components/SiteHeader.tsx
```diff
@@ -62,7 +62,9 @@ const SiteHeader = () => {
 
   const changeLanguage = () => {
     const newLang = LNG_CHANGE[i18n.language];
-    const path = (location.pathname + location.search).replace(`/${i18n.language}/`, `/${newLang}/`);
+    // Can't rely on there being a trailing slash at the base page (.e.g, `/en` and `/en`/ are both valid), and can't
+    // ensure project IDs don't begin with an `en` or `fr`. Thus, we use a RegExp for language changing in the URL.
```
SanjeevLakhwani, line 59: How would it be ensure that projectIDs don't being with a lang

SanjeevLakhwani (commented):
I really like the changes made! The use of icons and cta buttons is great!

src/js/components/SiteHeader.tsx
davidlougheed: regex with a `^` to mean beginning of the string.

SanjeevLakhwani (approved):
lgtm!

PR #249, opened 2025-01-24
fix: properly implement force_catalogue option

davidlougheed:


davidlougheed:
(to go with https://github.com/bento-platform/bento/pull/305, oops)

gsfk (approved):

PR #250, opened 2025-01-24
fix: config/overview sometimes not re-fetched on scope change

davidlougheed:
some issues with hook dependencies meant when a dataset without data was selected from a project card entry, for example, the overview would not refresh. this should fix the issue.


gsfk (approved):
Is it possible to have scopeSet `true` and configIsInvalid `false`? 

Anyway, works fine.

davidlougheed:
yes, it's in that state after loading config - the scope is set (project and dataset are specified), and the config is valid for that scope.

PR #251, opened 2025-02-04
Use config hook in beacon 

gsfk:
Beacon was no longer pulling config values for max query params on loading, breaking the UI for filters.


davidlougheed (approved):
lgtm, oops.

PR #253, opened 2025-02-07
Ignore scope in beacon network

gsfk:
Ignore any local scope in calls to beacons in beacon network. 

Although the "Beacon Network" tab is available only at instance-wide scope, it seems like instances with a single project and dataset can have scope present in redux, which broke some assumptions about scope in the beacon code.


davidlougheed (approved):
Lgtm. Sorry about the janky scope logic

PR #255, opened 2025-02-12
lint: enforce hook dependencies + CustomEmpty -> tsx

davidlougheed:
if hook dependencies are truly optional, an exception comment disabling eslint for the line can be added


gsfk (approved):
Thanks

PR #256, opened 2025-02-27
Pull search options from beacon 

gsfk:
Previously, the beacon tab in bento public has used the same discovery config settings as the Search tab to set metadata search options. Since Bento 18, beacon now knows its own search options, so these can be pulled from the beacon endpoint `/filtering_terms` instead. 

- makes Beacon more self-contained and closer to its intended use: 
  - it's now beacon's job to call katsu to figure out what the search options are, rather than leaving this to the web client
  - leaves open the possibility of filters that apply to other services (eg Takuan)
  - ideally the beacon UI should only need to communicate with beacon
- leaves open the possibility of working with non-Bento beacons

... and in general relying more on beacon `/filtering_terms` instead of direct calls to katsu will simplify the operation of the beacon network.

Also:
- made beacon config call conditional (it should have always been conditional)
- remapped some fields in beacon network filters (since they still follow "querySections" syntax)

### DIFF
--- a/src/js/features/beacon/beacon.store.ts
+++ b/src/js/features/beacon/beacon.store.ts
@@ -48,7 +48,6 @@ export const getBeaconFilters = createAsyncThunk<
 >(
   'beacon/getBeaconFilters',
   (_, { getState, rejectWithValue }) => {
-    const token = getState().auth.accessToken;
     const projectId = getState().metadata.selectedScope.scope.project;
     const datasetId = getState().metadata.selectedScope.scope.dataset;
 
@@ -69,9 +68,7 @@ export const makeBeaconQuery = createAsyncThunk<
   BeaconQueryPayload,
   { state: RootState; rejectValue: string }
 >('beacon/makeBeaconQuery', async (payload, { getState, rejectWithValue }) => {
-  const token = getState().auth.accessToken;
   const projectId = getState().metadata.selectedScope.scope.project;
-  const headers = makeAuthorizationHeader(token);
   return axios
     .post(scopedBeaconIndividualsUrl(projectId), payload, authorizedRequestConfig(getState()))
     .then((res) => res.data)
--- a/src/js/features/beacon/beacon.store.ts
+++ b/src/js/features/beacon/beacon.store.ts
@@ -58,7 +58,7 @@ export const getBeaconFilters = createAsyncThunk<
   },
   {
     condition(_, { getState }) {
-      return getState().metadata.selectedScope.scopeSet && getState().beacon.filtersStatus === RequestStatus.Idle;
+      return getState().metadata.selectedScope.scopeSet && getState().beacon.filtersStatus !== RequestStatus.Pending;
     },
   }
 );
--- a/src/js/features/beacon/beacon.store.ts
+++ b/src/js/features/beacon/beacon.store.ts
@@ -1,6 +1,5 @@
 import { createSlice, createAsyncThunk } from '@reduxjs/toolkit';
 import axios from 'axios';
-import { makeAuthorizationHeader } from 'bento-auth-js';
 import { authorizedRequestConfig } from '@/utils/requests';
 
 import { EMPTY_DISCOVERY_RESULTS } from '@/constants/searchConstants';
--- a/src/js/features/beacon/beacon.store.ts
+++ b/src/js/features/beacon/beacon.store.ts
@@ -1,6 +1,7 @@
 import { createSlice, createAsyncThunk } from '@reduxjs/toolkit';
 import axios from 'axios';
 import { makeAuthorizationHeader } from 'bento-auth-js';
+import { authorizedRequestConfig } from '@/utils/requests';
 
 import { EMPTY_DISCOVERY_RESULTS } from '@/constants/searchConstants';
 import { BEACON_INFO_ENDPOINT } from '@/features/beacon/constants';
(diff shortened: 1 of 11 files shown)


src/js/types/beacon.ts
```diff
@@ -162,6 +162,39 @@ export interface BeaconServiceInfo {
...
+    beaconId: string;
+  };
+  response?: {
+    filteringTerms: BeaconFilteringTermFromResponse[];
+  };
+}
+
+export interface BeaconFilteringTermFromResponse {
+  type: 'alphanumeric';
+  id: string; // something more here?
```
SanjeevLakhwani, line 181: is the comment helpful?

src/js/types/beacon.ts
```diff
@@ -162,6 +162,39 @@ export interface BeaconServiceInfo {
...
+  bento: {
+    section: string;
+  };
+  units?: string;
+}
+
+export type BeaconFilterForQuery = Omit<BeaconFilteringTermFromResponse, 'bento'>;
+
+export interface BeaconFilterSection {
+  section_title: BeaconFilteringTermFromResponse['bento']['section'];
```
SanjeevLakhwani, line 194: - camel case 
- BeaconFilteringTermFromResponse['bento']['section'] just evaluates to a string so should be just labelled as string

SanjeevLakhwani: if you really want to provide a relation you can add a comment

src/js/types/beacon.ts
```diff
@@ -162,6 +162,39 @@ export interface BeaconServiceInfo {
...
+export interface BeaconFilteringTermsResponse {
+  meta?: {
+    beaconId: string;
+  };
+  response?: {
+    filteringTerms: BeaconFilteringTermFromResponse[];
+  };
+}
+
+export interface BeaconFilteringTermFromResponse {
```
SanjeevLakhwani, line 179: Maybe just BeaconFilteringTerm?

src/js/features/beacon/beacon.store.ts
```diff
@@ -28,6 +40,30 @@ export const getBeaconConfig = createAsyncThunk<BeaconConfigResponse, void, { st
...
+>(
+  'beacon/getBeaconFilters',
+  (_, { getState, rejectWithValue }) => {
+    const token = getState().auth.accessToken;
+    const projectId = getState().metadata.selectedScope.scope.project;
+    const datasetId = getState().metadata.selectedScope.scope.dataset;
+    const headers = makeAuthorizationHeader(token);
+
+    return axios
+      .get(scopedBeaconFilteringTermsUrl(projectId, datasetId), { headers: headers as Record<string, string> })
```
SanjeevLakhwani, line 56: use authorizedRequestConfig in util/requests.js

src/js/features/beacon/beacon.store.ts
```diff
@@ -28,6 +40,30 @@ export const getBeaconConfig = createAsyncThunk<BeaconConfigResponse, void, { st
...
+      .then((res) => res.data)
+      .catch(printAPIError(rejectWithValue));
+  },
+  {
+    condition(_, { getState }) {
+      return getState().metadata.selectedScope.scopeSet;
```
SanjeevLakhwani, line 62: also include isFetchingFilters

SanjeevLakhwani (changes requested):

src/js/types/beacon.ts
gsfk: haha, probably not

src/js/types/beacon.ts
gsfk: If this is a comment on the naming I agree it's not great. But, owing to complications in the beacon spec, I need to distinguish between:
- "filtering terms" in the response from beacon endpoint `/filtering_terms`
- "filtering terms" in the beacon query

These are related but do not match exactly. But I'll look at these again and see if I can come up with something cleaner.

src/js/types/beacon.ts
gsfk: Fixed, although note that the snake case is borrowed from search.ts.

gsfk: ah, in fact the dependency doesn't really need to be there anymore, I can rewrite a few other types to get rid of this.

src/js/types/beacon.ts
gsfk: Kept types as they are but tried to making the naming a bit more clear

src/js/features/beacon/beacon.store.ts
```diff
@@ -58,6 +94,10 @@ const initialState: BeaconState = {
   isFetchingBeaconConfig: false,
   beaconAssemblyIds: [],
 
+  // filters
+  isFetchingFilters: false,
```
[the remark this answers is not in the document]
gsfk: huh, you guys snuck this in on me.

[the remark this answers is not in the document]
davidlougheed: yeah sorry - i'll try to roll it out across the rest of the site so that there isn't a confusing mix.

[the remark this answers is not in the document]
gsfk: I like it, although TS seems fussy about it.

src/js/features/beacon/beacon.store.ts
gsfk: done, although now uses `RequestStatus`

src/js/types/beacon.ts
gsfk: Can we change them all at once in a separate pr? 

If we change beacon section_title and regular search section_title at different times there's going to be a bunch of weird intermediate types needed (since the beacon network will need a type that's essentially `SectionButWithCamelCase`... which will be deprecated almost immediately when the fix is applied everywhere).

davidlougheed (approved):
lgtm

PR #257, opened 2025-03-18
feat: new search UI

davidlougheed:
requires Katsu to be `edge` / `develop` branch latest


src/js/components/Search/OrDelimiter.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+const lineStyle: CSSProperties = { width: 1, flex: 1, backgroundColor: '#f0f0f0' };
+
+const OrDelimiter = memo(() => {
+  const t = useTranslationFn();
+  return (
+    <Flex vertical={true} gap={8} align="center">
+      <div style={lineStyle}></div>
+      <div style={{ fontWeight: 'bold', color: '#595959' }}>{t('OR')}</div>
+      <div style={lineStyle}></div>
```
SanjeevLakhwani, line 13: Might be useful to use a divider with the vertical option instead of divs. If not the divs can be self closing. `<div ..../>`

src/js/components/Search/RequestStatusIcon.tsx
SanjeevLakhwani, line 1: Changes suggested in the notion doc

src/js/components/Search/SearchFilterInput.tsx
SanjeevLakhwani, line 1: suggested changes in notion doc

SanjeevLakhwani (changes requested):

src/js/components/Util/DefaultLayout.tsx
```diff
@@ -22,6 +22,7 @@ const DefaultLayout = () => {
             <Outlet />
           </Content>
           <SiteFooter />
+          <FloatButton.BackTop target={() => document.getElementById('content-layout')!} />
```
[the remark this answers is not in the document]
davidlougheed: i think aligned but with a different colour to emphasize chart management would be good.

src/js/components/Beacon/BeaconCommon/AssemblyIdSelect.tsx
```diff
@@ -8,7 +9,7 @@ const AssemblyIdSelect = ({ field, beaconAssemblyIds, disabled }: AssemblyIdSele
 
   return (
     <Form.Item name={field.name} label={t(field.name)} rules={field.rules}>
-      <Select style={{ width: '100%' }} disabled={disabled} options={assemblyIdOptions} />
+      <Select style={WIDTH_100P_STYLE} disabled={disabled} options={assemblyIdOptions} />
```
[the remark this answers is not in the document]
davidlougheed: done

src/js/components/Search/OrDelimiter.tsx
davidlougheed: ant has too many styles, ive factored this into a mini component tho

src/js/components/Search/SearchFilterInput.tsx
davidlougheed: i want it discoverable, and pushing it below the fold imo makes it less discoverable...

SanjeevLakhwani: we could maybe make start with something like create a search that asks to choose b/w filters or free text?

davidlougheed: ideally I want the min. # clicks to get to a FTS.

gsfk (commented):
Looks good so far, questions and a comment, but all of them minor:


1. are flat results better than results divided by project / dataset? (it looks like results are already returned in order by project / dataset rather than jumbled together, so that's good)
2. is "Manage Columns" going to get more columns?
3. is the results table going to outgrow this space it has now? Will we ever want a regular results table below the charts instead of forcing them to share?
4. for searches with no results, can we get a better message than "showing 1-0 of 0"?

But I think this is good as-is, except perhaps for the last point. 

Also - off topic and not really a UI issue - it would be nice to eventually allow filters and free text in the same search. This could be thrown into beacon I guess, although I was waiting for free text to speed up a bit (currently ~3 min on BQC19 data).

davidlougheed:
@gsfk 

1. I figure the scoping is better for deciding this? but we could maybe have a UI breakdown too, not sure.
2. yes, eventually
3. perhaps – I'm open to redesigns, but for now I wanted something that naturally fit into bento public
4. fixing this.

src/js/components/Provenance/Catalogue/CatalogueCard.tsx
```diff
@@ -157,7 +157,7 @@ const CatalogueCard = ({ project }: { project: Project }) => {
           datasets.length ? (
             <>
               <Title level={5} style={{ marginTop: 0 }}>
-                {t('Datasets')}
+                {t('entities.dataset', T_PLURAL_COUNT)}
```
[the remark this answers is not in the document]
davidlougheed: I wanted a common key for the plural and singular forms, which is why the key got changed. we should maybe change around the display, but for now I've preserved the old form ("Datasets")

gsfk:
Search still has [this bug](https://redmine.c3g-app.sd4h.ca/issues/2504) when changing scope, although perhaps that's out of scope for this PR

gsfk (commented):
Looks good except for the existing scope bug, not sure if that needs to be fixed here or not.

davidlougheed:
@gsfk working on a fix for that.

gsfk (approved):

src/js/components/Util/ErrorText.tsx
SanjeevLakhwani, line 1: could just be a css class for this?

src/js/components/Util/MonospaceText.tsx
SanjeevLakhwani, line 1: could just be a css class for this?

SanjeevLakhwani (changes requested):
Also to solve
- search param vanishing on tab switch
- specific search only (free text/filter) search params

src/js/components/Search/Search.tsx
```diff
@@ -84,70 +110,142 @@ const RoutedSearch = () => {
...
+              />
+              {queryDataPerm && (
+                // If we have the query:data permission on the current scope, we're allowed to run free-text searches on
+                // the data, so show the free-text search form:
+                <>
+                  <OrDelimiter vertical={!isSmallScreen} />
+                  <SearchFreeText
+                    focused={queryMode === 'text'}
+                    onFocus={onTextFocus}
+                    style={{ flex: 1, ...focusedStyle(queryMode === 'text') }}
```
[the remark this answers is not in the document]
davidlougheed: this doesn't work because of the OR element which isn't flex-1, but I'll use the `flex-1` class here instead.

src/js/components/Util/MonospaceText.tsx
davidlougheed: done.

gsfk:
Did we lose text search?

gsfk (approved):
Working fine for me. I may need to migrate my discovery config for new katsu, although that's presumably separate from this PR.

SanjeevLakhwani (approved):
lgtm

PR #260, opened 2025-03-21
Beacon filter improvements

gsfk:
Improvements to Beacon metadata search UI:
- Don't let users pick the same filter more than once: search fields that are already chosen are now disabled.
- Add helptext for each search option (reused `OptionDescription` component although I'd rather have ⓘ than the current circled question mark).
- Some code cleanup


src/js/components/Beacon/BeaconCommon/Filter.tsx
```diff
@@ -12,28 +12,34 @@ import type {
...
+}: FilterProps) => {
   const t = useTranslationFn();
 
   const [valueOptions, setValueOptions] = useState([{ label: '', value: '' }]);
 
-  const handleSelectKey = (_: unknown, option: GenericOptionType) => {
-    // set dropdown options for a particular key
-    // ie for key "sex", set options to "MALE", "FEMALE", etc
+  const handleSelectKey = (searchFieldId: string, option: GenericOptionType) => {
+    // update which search field this filter is using ("sex", "age", etc)     //naming confusion here, that "filter" means both the filters you can choose and the current options present in the form
```
davidlougheed, line 34: wrap to 120 chars ideally

src/js/components/Beacon/BeaconCommon/Filter.tsx
```diff
@@ -47,13 +53,21 @@ const Filter = ({ filter, form, beaconFiltersBySection, removeFilter, isRequired
   const renderLabel = (filter: BeaconFilterUiOptions) => {
```
davidlougheed, line 53: this is effectively a component now - maybe factor it out into a `FilterLabel` component

src/js/components/Beacon/BeaconCommon/Filters.tsx
```diff
@@ -29,37 +29,45 @@ const NetworkFilterToggle = () => {
...
-// ideally:
-// - should not permit you to make multiple queries on the same key (Redmine #1688)
-
 const BUTTON_STYLE = { margin: '10px 0' };
 
 const Filters = ({ filters, setFilters, form, beaconFiltersBySection, isNetworkQuery }: FiltersProps) => {
   const t = useTranslationFn();
+  const [filterIndex, setFilterIndex] = useState<number>(1);
 
   const { maxQueryParameters: maxFilters, maxQueryParametersRequired } = useConfig();
```
davidlougheed, line 38: katsu is supposed to return max int for max query params if the user has `query:data`, so I don't think this `maxQueryParametersRequired` state is actually needed

davidlougheed (changes requested):

src/js/components/Beacon/BeaconCommon/Filters.tsx
gsfk: No trouble to remove this, but is it needed in all the other places it's used?

davidlougheed: It's only used by Beacon as far as I can tell

src/js/components/Beacon/BeaconCommon/Filter.tsx
gsfk: ah, comment is dead, removed

davidlougheed (approved):
lgtm

PR #266, opened 2025-05-06
Fix/beacon filter bug

gsfk:
Bugfix for beacon filter handling:
- first filter was always ignored even if filled in
- previous logic for checking if you'd interacted with filters no longer applied

... let me know if I need `useCallback` here.

In the longer term we could probably replace this with the search component from the new UI, but I might wait for that to be merged first.


davidlougheed (approved):
lgtm

PR #267, opened 2025-05-07
refact(overview): singular entity ID counts/bools from katsu

davidlougheed:
* support bool "count" responses from Katsu

### DIFF
--- a/src/js/components/Overview/Counts.tsx
+++ b/src/js/components/Overview/Counts.tsx
@@ -29,7 +29,7 @@ const renderCount = (count: number | boolean | undefined, threshold: number): nu
     ? NO_RESULTS_DASHES
     : typeof count === 'boolean'
       ? count
-        ? `\u2265${threshold}`
+        ? `>${threshold}`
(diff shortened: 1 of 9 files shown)


gsfk (approved):
Work well, I'm wavering on whether a boolean UI response doesn't merit its own boolean-style format, although I don't have a suggestion that's not ugly (`true` or `exists` are both ugly).

I got this for a non-zero overview for a collection with a count threshold of 6:

`≥6`

Which seems like a good way of showing the certainty. But.... at least in beacon... the count thresholds are inclusive (results equal to the threshold are rounded to zero), so this should be 

`>6` 

(or `≥7` I guess).. it's entirely possible that katsu and beacon have diverged on their interpretation of threshold, and frankly I find it counterintuitive that it's inclusive, so it's probably worth reviewing: 
- whether katsu and beacon treat it the same, and 
- if that's really the way we want to treat it

davidlougheed:
you're right, it should be `>` not `≥`

gsfk (approved):

PR #272, opened 2025-05-27
fix(search): missing search translations

davidlougheed:


gsfk (commented):
Seems like filter options are not being translated.

The first option I'm offered for "Statut fonctionnel" is "I have no problems in walking about" instead of the expected "Je n’ai aucun problème à marcher". Beacon shows the French one, so it's not a translation config problem.

PR #274, opened 2025-05-29
deprecate old beacon overview response format

gsfk:
Deprecate old beacon overview response format. (Redmine #2541)

Beacon currently sends old and new formats together, so this is non-breaking.


davidlougheed (approved):
lgtm.

PR #275, opened 2025-05-30
Beacon uncensored zeros

gsfk:
Handle uncensored zero results from beacon: show "0" instead of "Insufficient data available". (Redmine #2360)

Related changes:
- remove stale beacon search overview values when no results (bug was masked by censored count handling)
- fix threshold handling (threshold is inclusive)
- remove prop drilling for `uncensoredCounts` value

... I'd hoped to eliminate one of the two bools responsible for showing zero or "insufficient data" (`hasInsufficientData` and `uncensoredCounts`) but the first one is a legacy of katsu's response for a censored zero response, which is just:

 `{message: "Insufficient data available."}`

with no actual zero to count.

### DIFF
--- a/src/js/components/Search/SearchResultsCounts.tsx
+++ b/src/js/components/Search/SearchResultsCounts.tsx
@@ -8,6 +8,7 @@ import CountsTitleWithHelp from '@/components/Util/CountsTitleWithHelp';
 import { COUNTS_FILL } from '@/constants/overviewConstants';
 import { NO_RESULTS_DASHES } from '@/features/search/constants';
 import { useHasScopePermission, useTranslationFn } from '@/hooks';
+import { useCanSeeUncensoredCounts } from '@/hooks/censorship';
 import type { SearchResultsUIPage } from '@/features/search/types';
 import type { DiscoveryResults, OptionalDiscoveryResults } from '@/types/data';
 import { RequestStatus } from '@/types/requests';
@@ -21,7 +22,6 @@ const SearchResultsCounts = ({
   selectedPage,
   setSelectedPage,
   hasInsufficientData,
-  uncensoredCounts,
   message,
 }: SearchResultsCountsProps) => {
   const t = useTranslationFn();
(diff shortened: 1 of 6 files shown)


src/js/components/Search/SearchResultsCounts.tsx
```diff
@@ -21,14 +22,14 @@ const SearchResultsCounts = ({
...
   hasInsufficientData,
-  uncensoredCounts,
   message,
 }: SearchResultsCountsProps) => {
   const t = useTranslationFn();
 
   const { individualCount, biosampleCount, experimentCount } = results;
   const { hasPermission: queryDataPerm } = useHasScopePermission(queryData);
   const individualsClickable = !!setSelectedPage && queryDataPerm;
+  const uncensoredCounts = useCanSeeUncensoredCounts();
```
davidlougheed, line 32: makes sense, thanks for removing the prop drilling

src/js/features/beacon/utils.ts
```diff
@@ -78,12 +78,18 @@ export const extractBeaconDiscoveryOverview = (response: BeaconQueryResponse): O
   // Bento-specific counts/chart data
   ...(response.info?.bento
     ? {
-        biosampleCount: response.info.bento.biosamples?.count,
+        biosampleCount: response?.info.bento.biosamples?.count,
```
davidlougheed, line 81: why this change? if response.info.bento exists, it follows that response exists.

src/js/hooks/censorship.ts
```diff
@@ -9,5 +9,5 @@ export const useCanSeeUncensoredCounts = () => {
   // Used mostly for UI - showing dashes vs "0".
   //  - If we have query:data permissions or the low cell count threshold is low enough that we get uncensored counts,
   //    then this becomes true.
-  return queryDataPerm || countThreshold <= 1;
+  return queryDataPerm || countThreshold <= 0;
```
davidlougheed, line 12: right... oops. can you fix the comment as well? "low enough" --> "0"

davidlougheed (changes requested):

davidlougheed (approved):

PR #276, opened 2025-06-03
feat(overview): collapsible overview sections

davidlougheed:
goal: allow users to reduce the possibly-overwhelming amount of info shown on the overview page without having to fully hide charts/sections via the chart manager.

also (newly) allows collapsing the overview datasets view.


gsfk (approved):
It took me a moment to even figure out what the feature was, which I think is good, since it doesn't add a lot of noise to the page.

PR #283, opened 2025-09-03
Fixed rule call sequence

SanjeevLakhwani:
Fixed the following bug: bento public behaves incorrectly when there are multiple projects. In particular it erroneously checks the top-level `/public_rules`  in situations where it should check `/public_rules/<project_id>` instead.


davidlougheed (changes requested):

davidlougheed (approved):
lgtm

PR #356, opened 2026-04-01
feat: general page header component + catalogue header

davidlougheed:
goal is to refactor moving towards a unified page header element for catalogue (replacing the basic AboutBox and splitting about content from the about box so the former can be reused) + sub-pages, bring Bento closer to the PCGL design, and clean up some confusingly-named components a bit.

in the long run, the new data catalogue + restyled portal will build on this work.

ticket: https://redmine.c3g-app.sd4h.ca/issues/2781


src/styles.css
```diff
@@ -190,32 +190,69 @@ body {
...
+  left: 0;
+  right: 0;
+  z-index: 20;
+}
+
+#page-header.catalogue-mode {
+  top: -1px;
+  padding-bottom: var(--content-padding-v);
+  background-color: var(--white-90);
+  box-shadow: inset 0 -1px 0 rgba(0, 0, 0, 0.05);
```
SanjeevLakhwani: this is barely visible

src/js/components/PageHeader.tsx
SanjeevLakhwani, line 1: Finalize the file

SanjeevLakhwani (changes requested):
Good addition.

Few nitpicks.
It is hard to differentiate or get a feel of what is the extra subheader is trying to convey. Some key differentiation like a different background or a stronger shadow could be useful to convey the direction of reading. As Sequentially it should be the second thing one might look after the logo (bento-public) but as it lacks a title (might not be the case everytime) it visually conflicts with the projects.

Image

src/js/components/PageHeader.tsx
SanjeevLakhwani, line 1: Add a loader component instead of the default -- for dataset counts. There is nothing to hit that we are waiting for the value os it seems that it is empty. This is important here cause the wait time is significat.

SanjeevLakhwani (changes requested):
Looks good. The design of the banner still needs to be finalized.

SanjeevLakhwani (commented):
The loader looks good. I would just like to get the teams consensus on the style now.

davidlougheed:
I might override any discussions on style and this would be a separate ticket. I don't have the bandwidth to balance that discussion and everything else given the PCGL research portal MVP timeline.

src/js/components/PageHeader.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const dispatch = useAppDispatch();
+  const { projectsStatus, datasetsByID } = useMetadata();
+  const renderCount = useRenderCount();
+  const { nodeCountsOrBools, nodeCountsOrBoolsFetched } = useSearchQuery();
+  const { scopeSet } = useSelectedScope();
+  const t = useTranslationFn();
+
+  useEffect(() => {
+    if (scopeSet && !nodeCountsOrBoolsFetched) dispatch(performKatsuDiscovery());
+  }, [dispatch, scopeSet, nodeCountsOrBoolsFetched]);
```
SanjeevLakhwani, line 36: Might be useful to separate out as a hook. I am not sure if this is the only call to it also that counts are included in the projects call also.

src/js/components/PageHeader.tsx
```diff
@@ -0,0 +1,113 @@
...
+            {t(`entities.${entity}`, { count }).toLocaleLowerCase()}
+          </li>
+        );
+      })}
+    </ul>
+  );
+};
+
+const PageHeader = ({ children, mode }: { children: ReactNode; mode: 'catalogue' | 'page' }) => {
+  const catalogue = mode === 'catalogue';
```
SanjeevLakhwani: Given that we only have two modes catalogue and page ("page" also is a bit ambiguos). Might be useful to just have catalogue as a boolean argument, for clarity and avoiding the two step boolean extraction.

src/js/components/PageHeader.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const renderCount = useRenderCount();
+  const { nodeCountsOrBools, nodeCountsOrBoolsFetched } = useSearchQuery();
+  const { scopeSet } = useSelectedScope();
+  const t = useTranslationFn();
+
+  useEffect(() => {
+    if (scopeSet && !nodeCountsOrBoolsFetched) dispatch(performKatsuDiscovery());
+  }, [dispatch, scopeSet, nodeCountsOrBoolsFetched]);
+
+  const nDatasets = Object.keys(datasetsByID).length;
```
SanjeevLakhwani: O(n) step, might be worth memo-izing

src/js/components/PageHeader.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+const PAGE_HEADER_CATALOGUE_STYLE: CSSProperties = {
+  background: CATALOGUE_HEADER_BACKGROUND,
+  color: CATALOGUE_HEADER_TEXT_COLOR,
+};
+
+const CataloguePageHeaderStats = () => {
+  const dispatch = useAppDispatch();
+  const { projectsStatus, datasetsByID } = useMetadata();
+  const renderCount = useRenderCount();
```
SanjeevLakhwani, line 28: Me no likey name, it seams like it is something that is calculating the number of times the page is being rendered.

src/js/components/PageHeader.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  // Effect for adding the data-stuck attribute to the page header when we scroll down.
+  useEffect(() => {
+    if (catalogue) return; // no sticky header in catalogue mode
+    const observer = new IntersectionObserver(
+      ([e]) => e.target.toggleAttribute('data-stuck', e.intersectionRatio < 1),
+      { threshold: [1], root: document.getElementById('content-layout') }
+    );
+
+    const st = document.getElementById('page-header');
```
SanjeevLakhwani: me no likey name

src/js/components/PageHeader.tsx
davidlougheed: this is consistent with several other uses in the codebase

src/js/components/PageHeader.tsx
davidlougheed: this specific effect is only used here. the dispatch of this action only occurs here and one other place. i added a comment here to make it a bit clearer what's going on.

src/js/components/PageHeader.tsx
SanjeevLakhwani: I am unfortunately most likely going to change this name. I understand that it is used at other places but I think it is worth changing (at every other place including). I can understand if this is out of scope. Could you maybe just add a comment in the origin of the hook file?

davidlougheed: added a todo

SanjeevLakhwani (approved):

PR #358, opened 2026-04-09
feat: range filter UI for numeric and date fields

SanjeevLakhwani:
## Summary

- Adds `RangeFilterInput` component with bracket toggles (`[`/`(` and `]`/`)`) and `InputNumber`/`DatePicker` bound inputs for numeric and date fields respectively
- Range inputs are only shown when the user has `query:data` permissions; unauthenticated users see the existing enum `Select` dropdown
- Range strings (`[x,y)`, `(x,y]`, etc.) are validated in the URL router and passed through to the backend as-is — no serialization changes needed
- Multi-value accumulation in URL validation is now guarded to string fields only (range fields take a single value)


gsfk (commented):
quick first pass: 

- both of them will launch a query with only min or only max, which creates a bad request
- both allow max < min
- I don't find the `[`,`(` options all that intuitive, for either option (in fact I've never seen this used for dates)
- date picker is broken, it can't change the month or year
- the choice of month is not related to the data (the date picker defaulted to today, but the bins returned by /discovery_search_fields for my data range from Jan 2020 to December 2023)

src/js/components/Search/RangeFilterInput.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const lowerDateValue = lowerStr ? dayjs(lowerStr, DATE_FORMAT) : null;
+  const upperDateValue = upperStr ? dayjs(upperStr, DATE_FORMAT) : null;
+
+  return (
+    <Space.Compact className="flex-1">
+      <Button onClick={onLowerBracketToggle}>{lowerOpen ? '(' : '['}</Button>
+      {isNumber ? (
+        <InputNumber
+          className="flex-1"
+          style={{ width: '100%' }}
```
davidlougheed: class `w-full`

src/js/components/Search/RangeFilterInput.tsx
```diff
@@ -0,0 +1,140 @@
...
+          min={numMin}
+          max={numMax}
+          value={lowerStr ? parseFloat(lowerStr) : null}
+          onChange={onLowerNumberChange}
+          placeholder="min"
+        />
+      ) : (
+        <DatePicker
+          className="flex-1"
+          style={{ width: '100%' }}
```
davidlougheed: idem

src/js/components/Search/RangeFilterInput.tsx
```diff
@@ -0,0 +1,140 @@
...
+          value={lowerDateValue?.isValid() ? lowerDateValue : null}
+          onChange={onLowerDateChange}
+          placeholder="start"
+        />
+      )}
+      <Button disabled>–</Button>
+      {isNumber ? (
+        <InputNumber
+          className="flex-1"
+          style={{ width: '100%' }}
```
davidlougheed: idem

src/js/components/Search/RangeFilterInput.tsx
```diff
@@ -0,0 +1,140 @@
...
+          min={numMin}
+          max={numMax}
+          value={upperStr ? parseFloat(upperStr) : null}
+          onChange={onUpperNumberChange}
+          placeholder="max"
+        />
+      ) : (
+        <DatePicker
+          className="flex-1"
+          style={{ width: '100%' }}
```
davidlougheed: idem

src/js/features/search/hooks.ts
```diff
@@ -21,6 +21,11 @@ export const useQueryFilterFields = (): QueryFilterField[] => {
   );
 };
 
+export const useSearchFilterFields = (): SearchFieldAndOptions[] => {
```
davidlougheed, line 16: confusingly similar name + functionality to the above hook - please consolidate them. I'm not convinced the above hook needs to exist at all, since `QueryFilterField` just seems to be `SearchFieldAndOptions` without the field definition, which isn't that useful. I would eliminate the former type entirely.

davidlougheed (changes requested):
first-pass code review

src/js/components/Search/RangeFilterInput.tsx
```diff
@@ -0,0 +1,185 @@
...
+        <DatePicker
+          className="flex-1 w-full"
+          format={DATE_FORMAT}
+          value={lowerDateValue?.isValid() ? lowerDateValue : null}
+          onChange={onLowerDateChange}
+          placeholder="start"
+          disabledDate={(d) => !!(upperDateValue?.isValid() && d.isAfter(upperDateValue, 'day'))}
+        />
+      )}
+      <Button disabled>–</Button>
```
davidlougheed: this shouldn't be semantically a button. I also think it ends up taking too much space visually; i can't even see a full date on my screen.

src/js/components/Search/RangeFilterInput.tsx
```diff
@@ -0,0 +1,185 @@
...
+      {isNumber ? (
+        <InputNumber
+          className="flex-1 w-full"
+          controls={false}
+          value={lowerStr ? parseFloat(lowerStr) : null}
+          onChange={onLowerNumberChange}
+          placeholder="min"
+        />
+      ) : (
+        <DatePicker
```
davidlougheed: is there an easy way to build in antd's `DatePicker.RangePicker` component, or is there a reason not to use it?

src/js/components/Search/RangeFilterInput.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+  const lowerNum = isNumber && lowerStr ? parseFloat(lowerStr) : null;
+  const upperNum = isNumber && upperStr ? parseFloat(upperStr) : null;
+  const boundsInverted = isNumber
+    ? lowerNum !== null && upperNum !== null && upperNum < lowerNum
+    : !!(lowerDateValue?.isValid() && upperDateValue?.isValid() && upperDateValue.isBefore(lowerDateValue, 'day'));
+
+  return (
+    <Space.Compact className="flex-1">
+      <Button onClick={onLowerBracketToggle}>{lowerOpen ? '(' : '['}</Button>
```
davidlougheed: for dates at least, i'd just make everything a closed interval to save space and make things a bit more intuitive. date ranges are almost always selected as closed intervals on other websites.

src/js/components/Search/RangeFilterInput.tsx
```diff
@@ -0,0 +1,185 @@
...
+  return `${lowerOpen ? '(' : '['}${lowerStr},${upperStr}${upperOpen ? ')' : ']'}`;
+};
+
+type RangeFilterInputProps = {
+  definition: NumberField | DateField;
+  value: FilterValue;
+  onChange: (v: FilterValue) => void;
+};
+
+const RangeFilterInput = ({ definition, value, onChange }: RangeFilterInputProps) => {
```
davidlougheed: if I change the field, the value isn't adequately removed. (for example if i go from a date with `2026-xxx` selected to a number, `2026` is preserved)

src/js/components/Search/RangeFilterInput.tsx
```diff
@@ -0,0 +1,185 @@
...
+  const upperNum = isNumber && upperStr ? parseFloat(upperStr) : null;
+  const boundsInverted = isNumber
+    ? lowerNum !== null && upperNum !== null && upperNum < lowerNum
+    : !!(lowerDateValue?.isValid() && upperDateValue?.isValid() && upperDateValue.isBefore(lowerDateValue, 'day'));
+
+  return (
+    <Space.Compact className="flex-1">
+      <Button onClick={onLowerBracketToggle}>{lowerOpen ? '(' : '['}</Button>
+      {isNumber ? (
+        <InputNumber
```
davidlougheed: we need to give some hints from the field definition on minimum/maximum values, since otherwise it's unclear what range to search in. 

the field info tooltips could also have information on the valid range for number fields.

e.g., i have no idea what to search range-wise here:

Image

davidlougheed (changes requested):

gsfk (commented):
More comments:
- click-to-search for dates is broken
- can it wait for you to stop typing before launching a search?
- if you have one filter open, and while editing a query you mange to produce a bad request, it reloads the results as if no filter is applied, which seems confusing, although arguably this is user error
- search config for numeric fields will sometimes have useful values for min and max values

src/js/components/Search/RangeFilterInput.tsx
SanjeevLakhwani: Semantically yes but for ease of visual organization when rendered it has visual consistency in the field. Also now dates are a range picker.

src/js/components/Search/NumberRangeFilterInput.tsx
```diff
@@ -0,0 +1,102 @@
...
+    <Space.Compact className="flex-1">
+      <Button onClick={onLowerBracketToggle}>{lowerOpen ? '(' : '['}</Button>
+      <InputNumber
+        className="flex-1 w-full"
+        controls={false}
+        value={lowerStr ? parseFloat(lowerStr) : null}
+        onChange={onLowerNumberChange}
+        placeholder={minimum != null ? String(minimum) : 'min'}
+      />
+      <Button disabled>–</Button>
```
davidlougheed: if this is left as a button, add `aria-hidden` prop and alter styling to not be exactly like a disabled button

* reduce horiz padding
* restore text color to be closer to black
* reset cursor to normal pointer

but in honesty i'd prefer a div or something.

src/js/components/Search/NumberRangeFilterInput.tsx
```diff
@@ -0,0 +1,102 @@
...
+    [lowerStr, lowerOpen, upperOpen, emitValue]
+  );
+
+  const lowerNum = lowerStr ? parseFloat(lowerStr) : null;
+  const upperNum = upperStr ? parseFloat(upperStr) : null;
+  const boundsInverted = lowerNum !== null && upperNum !== null && upperNum < lowerNum;
+
+  return (
+    <Space.Compact className="flex-1">
+      <Button onClick={onLowerBracketToggle}>{lowerOpen ? '(' : '['}</Button>
```
davidlougheed: add tooltip showing meaning of this (inclusive/exclusive). same for other end of the range.

src/js/components/Search/NumberRangeFilterInput.tsx
```diff
@@ -0,0 +1,102 @@
...
+      <InputNumber
+        className="flex-1 w-full"
+        controls={false}
+        value={lowerStr ? parseFloat(lowerStr) : null}
+        onChange={onLowerNumberChange}
+        placeholder={minimum != null ? String(minimum) : 'min'}
```
davidlougheed: use `!==`

src/js/components/Search/NumberRangeFilterInput.tsx
```diff
@@ -0,0 +1,102 @@
...
+      />
+      <Button disabled>–</Button>
+      <InputNumber
+        className="flex-1 w-full"
+        controls={false}
+        status={boundsInverted ? 'error' : undefined}
+        value={upperStr ? parseFloat(upperStr) : null}
+        onChange={onUpperNumberChange}
+        placeholder={maximum != null ? String(maximum) : 'max'}
```
davidlougheed: use `!==`

src/js/components/Search/NumberRangeFilterInput.tsx
```diff
@@ -0,0 +1,102 @@
...
+  // Render-time derived-state sync (React recommended pattern for props → state).
+  // Only sync when rawValue changed AND the change came from outside (not from our own onChange call).
+  if (rawValue !== prevRawValueRef.current) {
+    prevRawValueRef.current = rawValue;
+    if (rawValue !== lastEmittedRef.current) {
+      setRangeState(parseBrackets(rawValue));
+    }
+  }
+
+  const emitValue = useCallback(
```
davidlougheed, line 46: throttle calls to this function to avoid spamming the search endpoint

once we (eventually) make search faster, we can tweak the throttle rate.

davidlougheed (changes requested):

src/js/components/Search/DateRangeFilterInput.tsx
```diff
@@ -0,0 +1,73 @@
...
+      emitValue(lStr, uStr);
+    },
+    [emitValue]
+  );
+
+  const lowerDateValue = lowerStr ? dayjs(lowerStr, DATE_FORMAT) : null;
+  const upperDateValue = upperStr ? dayjs(upperStr, DATE_FORMAT) : null;
+
+  return (
+    <DatePicker.RangePicker
```
davidlougheed, line 64: defaults to the current date, which isn't very useful given we know the start/end for dates usually

src/js/components/Search/NumberRangeFilterInput.tsx
```diff
@@ -0,0 +1,129 @@
...
+  const boundsInverted = lowerNum !== null && upperNum !== null && upperNum < lowerNum;
+
+  return (
+    <Space.Compact className="flex-1">
+      <Tooltip
+        title={`${t(lowerOpen ? 'search.range.lower_exclusive' : 'search.range.lower_inclusive')} — ${t('search.range.click_to_switch')}`}
+      >
+        <Button onClick={onLowerBracketToggle}>{lowerOpen ? '(' : '['}</Button>
+      </Tooltip>
+      <InputNumber
```
davidlougheed, line 108: enforce actual minimum if minimum doesnt equal taper left

davidlougheed (changes requested):

gsfk (changes requested):
Looking good, a couple remaining issues: 

- I can still produce a bad request this way: click-to-search on a numeric field, then delete the min value
- missing translations for "Start date" and "End date", the antd date picker needs special locale handling

gsfk (approved):

PR #360, opened 2026-04-15
style: tweak margins of search results table card

davidlougheed:
* reduces margin below pagination in search table card
* makes vertical margins consistent (12px, 1/2 of horiz margins) between top of card, top card bar with title/buttons, bottom of table to pagination, and pagination to end of card


SanjeevLakhwani (approved):
lgtm

PR #361, opened 2026-04-15
feat(clinphen): add biosample popovers for pre-click detail

davidlougheed:
adds popovers to biosample IDs in individual search table, experiment search table, and experiment table in the phenopacket view.

the goal here is to show info about a biosample to the user before they click on it, so they can know if they care before they get navigated away.

most of the way there; need to come up with some final names for components + perhaps tweak data re-use inside the phenopacket view.


SanjeevLakhwani (approved):
lgtm

PR #363, opened 2026-04-22
remove stale pagination params on change

gsfk:
Bugfix for Redmine #2794

Query params are now stored as array of arrays instead of object, so the expected pattern of spreading the old params and adding new ones no longer works (this just gives you the same param twice with two different values). 

We could revisit objects vs nested arrays at some point, for now here is a one-line fix for broken pagination params,

### DIFF
--- a/src/js/components/Search/SearchResultsTablePage.tsx
+++ b/src/js/components/Search/SearchResultsTablePage.tsx
@@ -30,7 +30,11 @@ import type {
 import { BentoRoute } from '@/types/routes';
 
 import { scopeSelectionEqual } from '@/features/metadata/utils';
-import { bentoKatsuEntityToResultsDataEntity, buildQueryParamsUrl } from '@/features/search/utils';
+import {
+  bentoKatsuEntityToResultsDataEntity,
+  buildQueryParamsUrl,
+  queryParamsWithoutKey,
+} from '@/features/search/utils';
 import { setEquals } from '@/utils/sets';
 
 import DatasetProvenanceModal from '@/components/Provenance/DatasetProvenanceModal';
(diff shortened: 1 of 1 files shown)


src/js/components/Search/SearchResultsTablePage.tsx
```diff
@@ -400,7 +400,11 @@ const SearchResultsTable = <T extends ViewableDiscoveryMatchObject>({
               }
               navigateToSameScopeUrl(
                 buildQueryParamsUrl(BentoRoute.Overview, [
-                  ...allQueryParams,
+                  //remove stale params
+                  ...allQueryParams.filter(
+                    (p) => p.length > 0 && !(p[0] == TABLE_PAGE_QUERY_PARAM || p[0] == TABLE_PAGE_SIZE_QUERY_PARAM)
+                  ),
+
```
SanjeevLakhwani: Writing it the following way is more readable:     `...queryParamsWithoutKey(allQueryParams, [TABLE_PAGE_QUERY_PARAM, TABLE_PAGE_SIZE_QUERY_PARAM]),
`

gsfk: So, a general function that replaces arbitrary stale params with new ones?

SanjeevLakhwani: Oh the function already exists, that's why

SanjeevLakhwani: Otherwise this would have been fine

gsfk: ah cheers

SanjeevLakhwani (approved):
Dog LGTM

PR #364, opened 2026-04-30
feat: migrate to DatasetV2 with new provenance view

SanjeevLakhwani:
- Add TypeScript types for `DatasetV2` (`ProjectScopedDatasetModel`) and migrate `Project` store/type from `datasets` to `datasets_v2`
- Build new `DatasetV2Provenance` component replacing legacy DATS-based provenance tables; remove legacy table components and `DownloadDats`
- Update all consumers (routing, hooks, scope picker, overview, search, catalogue card, provenance modal) to use `datasets_v2`; re-fetch projects with `Accept-Language` header on language change


src/js/types/dataset.ts
```diff
@@ -0,0 +1,323 @@
...
+  | 'Data Provider'
+  | 'Data Controller'
+  | 'Data Processor'
+  | 'Data Contributor'
+  | 'Data Custodian'
+  | 'Data Producer'
+  | 'Partner'
+  | 'Stakeholder'
+  | 'Community Representative'
+  | 'Other';
```
gsfk, line 56: Comparing this with the model in bento_lib, is there a reason "Author", and "Corresponding Author" are missing?

gsfk (commented):
Working as expected, one inline question

src/js/types/datasetV2.ts
```diff
@@ -0,0 +1,323 @@
...
+  /** UUID of the project this dataset belongs to. */
+  project: string;
+
+  title: string;
+  description: string;
+  long_description?: LongDescription | null;
+
+  /** Taxonomic scope of the dataset. */
+  taxa?: (OntologyClass | string)[] | null;
+  keywords?: (string | OntologyClass)[] | null;
```
davidlougheed: use consistent ordering for these + maybe just make a small type called `StringOrOntologyClass`

src/js/components/Provenance/DatasetV2Provenance.tsx
```diff
@@ -0,0 +1,485 @@
...
+export const DatasetV2ProvenanceContent = ({ dataset }: { dataset: DatasetV2 }) => {
+  const t = useTranslationFn();
+
+  const keywords = dataset.keywords ?? [];
+  const taxa = dataset.taxa ?? [];
+  const stakeholders = dataset.stakeholders ?? [];
+  const publications = dataset.publications ?? [];
+  const fundingSources = Array.isArray(dataset.funding_sources) ? dataset.funding_sources : [];
+  const criteria = dataset.participant_criteria ?? [];
+  const counts = dataset.counts ?? [];
```
gsfk: shouldn't `dataset.counts` be `dataset.counts_by_entity`? My v2 datasets all have counts = null, and I never see any counts in the Provenance display.

gsfk (changes requested):
~~Looks like counts are not shown in provenance.~~

src/js/components/Overview/OverviewDatasets.tsx
```diff
@@ -23,7 +23,7 @@ const OverviewDatasets = ({
               dataset={d}
               format="card"
               filteredCounts={
-                countsByDataset ? (countsByDataset[d.identifier] ?? EMPTY_KATSU_ENTITY_COUNTS) : undefined
+                countsByDataset ? (countsByDataset[d.identifier] ?? EMPTY_KATSU_ENTITY_COUNTS) : d.counts_by_entity
```
davidlougheed, line 26: why this change? historically if this was undefined it would indicate to the component that no filtering had happened

src/js/components/Provenance/Catalogue/CatalogueCard.tsx
```diff
@@ -49,10 +49,10 @@ const CatalogueCard = ({ project }: { project: Project }) => {
 
   const isSmallScreen = useSmallScreen();
 
-  const { datasets, created, updated, title, description, identifier, counts } = project;
+  const { datasets_v2: datasets, created, updated, title, description, identifier, counts } = project;
 
   const { selectedKeywords, extraKeywords, extraKeywordCount } = useMemo(() => {
-    const keywords = datasets.flatMap((d) => d.dats_file.keywords ?? []).map((k) => t(k.value as string));
+    const keywords = datasets.flatMap((d) => d.keywords ?? []).map((k) => t(typeof k === 'string' ? k : k.label));
```
davidlougheed, line 56: comment: for next version, we will need to render these properly as ontology classes

src/js/components/Provenance/Catalogue/CatalogueCard.tsx
```diff
@@ -161,7 +161,13 @@ const CatalogueCard = ({ project }: { project: Project }) => {
                 }}
               >
                 {datasets.map((d) => (
-                  <Dataset parentProjectID={identifier} key={d.identifier} dataset={d} format="carousel" />
+                  <Dataset
+                    parentProjectID={identifier}
+                    key={d.identifier}
+                    dataset={d}
+                    format="carousel"
+                    filteredCounts={d.counts_by_entity}
```
davidlougheed, line 170: same comment as above

src/js/components/Provenance/DatasetV2Provenance.tsx
```diff
@@ -0,0 +1,485 @@
+import { Card, Descriptions, Flex, Tag, Typography } from 'antd';
```
davidlougheed: not a huge fan of having a bunch of components in the same file now, whereas we used to have some of the tables split out into multiple files.

src/js/types/metadata.ts
```diff
@@ -1,27 +1,18 @@
...
 import type { KatsuEntityCountsOrBooleans } from '@/types/entities';
+import type { DatasetV2 } from '@/types/datasetV2';
 
 export interface Project {
   identifier: string;
   title: string;
   description: string;
   discovery: DiscoveryConfig | null;
-  datasets: Dataset[];
+  datasets_v2: DatasetV2[];
```
davidlougheed: given youre removing the original dataset type, does this still need to be called v2?

davidlougheed (commented):
first pass.

in addition, i suspect, but haven't yet confirmed, that there are some missing/unstructured translations here that should be put in and structured.

src/js/types/metadata.ts
SanjeevLakhwani: I did that change just so that I don't miss any changes. I can rename it now

src/js/components/Provenance/DatasetV2Provenance.tsx
SanjeevLakhwani: just wanted to have a quick thing before the redesign pr

gsfk (commented):
Appears to work as expected, although I am coming across existing issues more than ever. 

See in particular Redmine 2742: when I click a dataset I sometimes get results for the entire project.

I'm also getting the converse problem: sometimes when exploring a dataset and then clicking "back to project" the project page gives me counts only for the dataset I was just looking at. Other datasets in the project show no counts at all. 

In both cases the scope is lagging behind what it should be. The first issue at least exists already in main, although possibly the projects page renders slower now, which makes the bug easier to produce.

src/js/components/Provenance/Dataset.tsx
```diff
@@ -99,7 +94,7 @@ const Dataset = ({
       <Card
         title={<SmallChartCardTitle title={title} />}
         size="small"
-        className={clsx('shadow', 'h-full', { 'opacity-50': faded })}
```
davidlougheed, line 102: why kill the faded out on search thing? this was intended as a feature to highlight searches that eliminate certain datasets.

src/js/types/dataset.ts
```diff
@@ -0,0 +1,322 @@
...
+  url: string;
+  theme: 'light' | 'dark' | 'default';
+  description?: string | null;
+  /** Whether the logo contains branding text beside the logo image. */
+  contains_text: boolean;
+}
+
+// ---- Spatial coverage (GeoJSON Feature) ----
+
+type GeoJSONGeometry =
```
davidlougheed: is there an available typescript type package for this?

davidlougheed (changes requested):

src/js/types/dataset.ts
SanjeevLakhwani: Do you want me to use the package for this?

davidlougheed: is there not already something used by the map view?

src/js/components/Provenance/Dataset.tsx
SanjeevLakhwani: it was tied to filtered/effective counts which was removed, I am gonna add it back.

gsfk (approved):
Approving since out of issues that are not already present in main or seemingly unrelated.

davidlougheed (approved):
looks good, should be merged alongside the other two PRs

PR #365, opened 2026-05-12
chore: hide last ingestion info in catalogue mode

davidlougheed:
as discussed in slack - hide last ingestion info when we're not in a single-project node.


gsfk (approved):
Works fine, this seems like the sort of thing that doesn't need to be shown quite so prominently.

PR #371, opened 2026-05-25
feat: add overridable instance-specific CSS file

davidlougheed:
* adds overridable instance-specific CSS file
* adds CSS-variable-based theming


SanjeevLakhwani (changes requested):
Following from this: https://5x.ant.design/docs/react/css-variables Given that we are already creating an instance css we can use css-vars for ant overrides also which would make this whole pr a one liner.

README.md
v-rocheleau, line 1: **question:** for convenience with containers, will we be able to mount a volume at a certain path to define the CSS class for a given instance?

PR #375, opened 2026-05-26
feat: show biosample detail directly with single biosample

davidlougheed:


gsfk (changes requested):
hmmm... doesn't actually work as advertised for me. Does this bypass the url handling for `expanded`?

I now get "Biosample" instead of "Biosamples (1)" but the row is still unexpanded: 

Image

gsfk (approved):

PR #377, opened 2026-05-28
fix biosample count total in beacon network results

gsfk:
Fix biosample counts for beacon network results, it was erroneously just repeating the experiment counts.


davidlougheed (approved):
lgtm

PR #383, opened 2026-06-05
fix: enforce min/max on all numeric range filter inputs

SanjeevLakhwani:
## Summary

- Fixes [#2817](https://redmine.c3g-app.sd4h.ca/issues/2817)
- `ManualBinsNumberFieldConfig` fields never enforced `min` on the lower input, the previous check gated on `taper_left` which only exists on `AutoBinsNumberFieldConfig` (age fields)
- No numeric field type enforced `max` on the upper input at all
- Now both `min` and `max` are applied to their respective inputs whenever the field config defines `minimum`/`maximum`


gsfk (commented):
Not ready to approve, since it's not immediately clear how to deal with max age. A max age of e.g. 110 is fine for bin config, but weird for search. The expected fix of just removing maximum from age gives a discovery config validation error. So I don't think this is a UI fix alone.

gsfk (approved):

PR #386, opened 2026-06-07
feat: catalogue UI redesign - faceted search, insights charts, compact cards

SanjeevLakhwani:
## Summary
Redmine feature ticket: [2688](https://redmine.c3g-app.sd4h.ca/issues/2688)
- Replaces the flat dataset grid with a faceted-search layout: sticky left rail with collapsible filter sections (Project, Data type, Taxa, Access, Data use, Status, Keywords), toolbar with search/sort/view-toggle on one line, and active-filter pill row with a pill-shaped "Clear all"
- Adds a Dataset Insights panel with SVG donut charts (By Status, By Project) and CSS bar charts (By Data type, By Keyword: top 5), all wired to the same facet state for two-way filtering
- Compact card redesign: status pill (title-case, Unassigned fallback), inline project pill with colored dot, keyword tags with primary-color border, counts + actions pinned to bottom
- Centralises project color assignment in Redux so card dots and insights chart use identical palette-cycled colors (lexicographic sort)
- Normalises `study_status` to title case (`Ongoing`, `Completed`) everywhere; datasets with no recognised status get `Unassigned`
- Syncs all filter state (`q`, `sort`, `view`, facet selections) to URL search params so searches are shareable and survive reload
- Removes `assays` facet (backed by `extra_properties`, not a typed dataset field); renames facet IDs `programs→projects`, `organisms→taxa` to match their backing fields

## Known Issues
- Charts and their theming will be consolidated after this PR. They would be moved to bento charts.
- In catalogue view, it has little customizability on filters, and which charts can be displayed. Would require work on discovery config and katsu to fully realize its potential.


src/js/components/Provenance/DatasetProvenance.tsx
```diff
@@ -365,20 +382,20 @@ export const DatasetProvenanceContent = ({ dataset }: { dataset: Dataset }) => {
           ) : null}
           {taxa.length > 0 && (
             <Item span={24} label={<DescLabel title={t('Taxa')} />}>
-              {taxa.map((k, i) => (
-                <Tag key={i} color="geekblue">
-                  {t(keywordLabel(k))}
-                </Tag>
-              ))}
+              <div style={{ display: 'flex', flexWrap: 'wrap', gap: 4 }}>
```
davidlougheed: use antd flex

src/js/components/Provenance/DatasetProvenance.tsx
```diff
@@ -365,20 +382,20 @@ export const DatasetProvenanceContent = ({ dataset }: { dataset: Dataset }) => {
...
             </Item>
           )}
           {keywords.length > 0 && (
             <Item span={24} label={<DescLabel title={t('Keywords')} />}>
-              {keywords.map((k, i) => (
-                <Tag key={i} color="cyan">
-                  {t(keywordLabel(k))}
-                </Tag>
-              ))}
+              <div style={{ display: 'flex', flexWrap: 'wrap', gap: 4 }}>
```
davidlougheed: idem

src/js/components/Scope/ScopeBreadcrumb.tsx
```diff
@@ -0,0 +1,49 @@
...
+import ScopedTitle from '@/components/Scope/ScopedTitle';
+
+type ScopeBreadcrumbProps = {
+  showSidebarToggle: boolean;
+  sidebarOverlayShown: boolean;
+  onToggleSidebar: () => void;
+  breadcrumbItems: BreadcrumbItemType[];
+};
+
+const ScopeBreadcrumb = ({ showSidebarToggle, sidebarOverlayShown, onToggleSidebar, breadcrumbItems }: ScopeBreadcrumbProps) => {
```
davidlougheed: can you please restore the pageheader component instead of changing this - out of scope.

src/public/assets/banner-bg.png
davidlougheed, line 1: duplicate file - please purge from git history

src/public/locales/en/default_translation_en.json
```diff
@@ -163,6 +163,36 @@
   "Created": "Created",
   "Updated": "Updated",
   "Back to catalogue": "Back to catalogue",
+  "Filters": "Filters",
```
davidlougheed: prefer structured translation keys

src/public/locales/en/default_translation_en.json
```diff
@@ -163,6 +163,36 @@
...
+  "Project": "Project",
+  "Data type": "Data type",
+  "Taxa": "Taxa",
+  "Data use": "Data use",
+  "Status": "Status",
+  "Keyword": "Keyword",
+  "Datasets": "Datasets",
```
davidlougheed: duplicate key

src/js/components/Provenance/Catalogue/Charts/DonutChart.tsx
```diff
@@ -0,0 +1,103 @@
+import { Card, Typography } from 'antd';
+import { useTranslationFn } from '@/hooks';
+import type { FacetId } from '@/features/catalogue/catalogue.store';
+import { COLOR_CHART_FALLBACK, COLOR_DONUT_TRACK } from '../constants';
+
+const { Text } = Typography;
+
+const DEFAULT_COLOR = COLOR_CHART_FALLBACK;
+const fmt = (n: number) => n.toLocaleString('en-US');
+const SIZE = 116,
```
davidlougheed, line 9: build components to be parametrized/reusable by default

src/js/components/Provenance/Catalogue/Catalogue.tsx
```diff
@@ -1,32 +1,95 @@
...
+  );
+
+  const { filtered, facetOptions } = useCatalogueFilter(allDatasets);
+
+  useEffect(() => {
+    const names = [...new Set(allDatasets.map(({ project }) => project.title))];
+    dispatch(setProjectColors(assignColors(names)));
+  }, [allDatasets, dispatch]);
+
+  const gridStyle: React.CSSProperties =
```
davidlougheed: use type imports instead of globals

src/styles.css
```diff
@@ -506,6 +496,351 @@ ul#page-header-stats li.loading {
...
+.catalogue-section-label {
+  font-size: 13px;
+  font-weight: 700;
+  text-transform: uppercase;
+  letter-spacing: 0.06em;
+  color: var(--cat-text-secondary);
+  white-space: nowrap;
+}
+
+.max-w-catalogue { max-width: 1240px; }
```
davidlougheed: this is too limiting. i'd prefer the 1400px max width from the rest of the site. there's already unused space on my macbook, where i could fit 4 cards instead of 3

src/js/components/Provenance/Dataset.tsx
```diff
@@ -97,38 +110,115 @@ const Dataset = ({
...
+              WebkitBoxOrient: 'vertical',
+              overflow: 'hidden',
+            }}
+          >
+            {t(description)}
+          </Text>
+        )}
+
+        {keywords.length > 0 && (
+          <div style={{ marginTop: 8, display: 'flex', flexWrap: 'wrap', gap: 4 }}>
```
davidlougheed: use antd flex + we should have a single component for keyword display at this point, it's done a lot.

src/js/components/Provenance/Dataset.tsx
```diff
@@ -97,38 +110,115 @@ const Dataset = ({
...
+            {keywords.map((kw) => (
+              <Tag
+                key={kw}
+                style={{
```
davidlougheed: duplicated styles from css class?

src/js/components/Provenance/Dataset.tsx
```diff
@@ -97,38 +110,115 @@ const Dataset = ({
...
+          border: `1px solid ${COLOR_BORDER}`,
+          boxShadow: SHADOW_CARD,
+          height: '100%',
+          display: 'flex',
+          flexDirection: 'column',
+          transition: 'box-shadow 0.2s, border-color 0.2s',
+          opacity: faded ? 0.5 : undefined,
+        }}
+        styles={{ body: { display: 'flex', flexDirection: 'column', height: '100%', padding: 16, gap: 0 } }}
+        onMouseEnter={(e) => {
```
davidlougheed: shouldn't be done with js

src/js/components/Provenance/Dataset.tsx
```diff
@@ -97,38 +110,115 @@ const Dataset = ({
...
-        className={clsx('shadow', 'h-full', { 'opacity-50': faded })}
-        style={{ minHeight: 200 }}
-        styles={{ body: { padding: '12px 16px', height: 'calc(100% - 53px)' } }}
-        extra={
-          <Button icon={<SolutionOutlined />} onClick={openProvenanceModal}>
-            {t('Provenance')}
-            <ExpandAltOutlined />
-          </Button>
-        }
+        style={{
```
davidlougheed: should be css

davidlougheed (changes requested):
i know you didn't request review yet, but flagging a few things here before you get to the review stage:

- i would have preferred if this was done in chunks (as outlined in the contributor guide draft - small PRs). dataset-only cards first, then faceting + search. this pr is fairly long as-is.
- the PCGL background is now in the repository twice (and now will be in the git history twice)
- other things as noted

src/js/components/Provenance/Catalogue/CatalogueBanner.tsx
```diff
@@ -0,0 +1,91 @@
...
+      {value} {label}
+    </Text>
+  </Space>
+);
+
+interface CatalogueBannerProps {
+  filteredDatasets: DatasetWithProject[];
+}
+
+const fmt = (n: number) => n.toLocaleString('en-US');
```
davidlougheed: use user locale

src/js/components/Scope/ScopeBreadcrumb.tsx
SanjeevLakhwani: It is not, page header had 2 purposes not linked to each other.

src/js/components/Provenance/Catalogue/Charts/BarChart.tsx
```diff
@@ -0,0 +1,65 @@
+import { Card, Typography } from 'antd';
+import { useTranslationFn } from '@/hooks';
+import type { FacetId } from '@/features/catalogue/catalogue.store';
+import { COLOR_BAR_DEFAULT } from '../constants';
+
+const { Text } = Typography;
+
+const DEFAULT_COLOR = COLOR_BAR_DEFAULT;
+const fmt = (n: number) => n.toLocaleString('en-US');
```
davidlougheed: use user locale

src/js/components/Provenance/Catalogue/Charts/DonutChart.tsx
```diff
@@ -0,0 +1,105 @@
+import { Card, Typography } from 'antd';
+import { useTranslationFn } from '@/hooks';
+import type { FacetId } from '@/features/catalogue/catalogue.store';
+import { COLOR_CHART_FALLBACK, COLOR_DONUT_TRACK } from '../constants';
+
+const { Text } = Typography;
+
+const DEFAULT_COLOR = COLOR_CHART_FALLBACK;
+const fmt = (n: number) => n.toLocaleString('en-US');
```
davidlougheed: idem

src/js/components/Provenance/Catalogue/constants.ts
```diff
@@ -0,0 +1,54 @@
+// Primary brand colour (PCGL dark blue)
```
davidlougheed: are all of these variables still used?, e.g., COLOR_INSIGHTS_BG

src/js/features/catalogue/useCatalogueUrlSync.ts
```diff
@@ -0,0 +1,61 @@
...
+  const dispatch = useAppDispatch();
+  const state = useCatalogueState();
+  const [searchParams, setSearchParams] = useSearchParams();
+  const skipFirstSync = useRef(true);
+
+  // Mount: read URL → hydrate Redux
+  useEffect(() => {
+    const rawSort = searchParams.get('sort');
+    const rawView = searchParams.get('view');
+    const sets: CatalogueFilterSets = {
```
davidlougheed: build from facet ids perhaps

src/js/components/Provenance/Catalogue/Catalogue.tsx
```diff
@@ -1,32 +1,101 @@
...
+  );
+
+  const { filtered, facetOptions } = useCatalogueFilter(allDatasets);
+
+  useEffect(() => {
+    const names = [...new Set(allDatasets.map(({ project }) => project.title))];
+    dispatch(setProjectColors(assignColors(names)));
+  }, [allDatasets, dispatch]);
+
+  const gridStyle: CSSProperties =
```
davidlougheed: could this be consolidated to always display using `grid` with just template columns changing?

src/js/features/catalogue/hooks.ts
```diff
@@ -0,0 +1,131 @@
...
+        if (selected.length === 0) continue;
+        const vals = getDatasetFacetValues(item)[fid];
+        if (!vals.some((v) => selected.includes(v))) return false;
+      }
+      return true;
+    }
+
+    const filtered = items.filter((item) => matchesItem(item, null));
+
+    const sortedFiltered = [...filtered].sort((a, b) => {
```
davidlougheed: imo this would be cleaner/more maintainable if you had a constant map of `Record<SortOption, DatasetSortFn>` and just did the lookup for the sort function using the sort option. this would also be (marginally) more computationally efficient.

src/js/components/Scope/ScopeBreadcrumb.tsx
```diff
@@ -0,0 +1,51 @@
...
+import ScopedTitle from '@/components/Scope/ScopedTitle';
+
+type ScopeBreadcrumbProps = {
+  showSidebarToggle: boolean;
+  sidebarOverlayShown: boolean;
+  onToggleSidebar: () => void;
+  breadcrumbItems: BreadcrumbItemType[];
+};
+
+const ScopeBreadcrumb = ({
```
davidlougheed: restore previous name since this becomes confusing with `breadcrumb -> scopedTitle -> breadcrumb`

also would help with reviewing to reduce css diff

gsfk (commented):
Looks nice, better than the usual antd default style. Some comments: 

- I assume you're going to get rid of `_devPcglOverride` before merging
- Also, I assume the charts will be moved to bento_charts eventually 
- There are two buttons to toggle the display. Does it need to be two or can it be a single button? 
- "status" isn't really something we're using outside of PCGL, so it probably doesn't need to appear in non-PCGL mode (currently it's a chart and a tag on the datasets)
- The cursor changes to a pointing finger over dataset counts but clicking does nothing
- we could consider having a (customizable?) banner background image for non-PCGL mode, I don't think Bento needs to look quite so austere 
- The filters strip on the left is quite a bit different from the filters strip found on the project or dataset page, even the font is different. Should they be more similar? 
- pretty sure if you hammer hard enough on Recharts, you can get this kind of donut-with-centre-text, although from what I remember the recharts version is far uglier.

src/styles.css
```diff
@@ -24,6 +24,32 @@
 
   --antd-blue-5: #1677ff;
   --antd-gray-7: #595959;
+
+  /* catalogue tokens */
+  --cat-primary: #054A74;
```
davidlougheed: maybe this can use primary colour from antd js?

src/public/locales/en/default_translation_en.json
```diff
@@ -533,7 +533,58 @@
...
+    },
+    "status": {
+      "ongoing": "Ongoing",
+      "completed": "Completed",
+      "unassigned": "Unassigned"
+    }
+  },
   "pcgl": {
+    "banner": {
+      "eyebrow": "PAN-CANADIAN GENOME LIBRARY",
```
davidlougheed, line 584: use normal case + text-transform instead

src/styles.css
```diff
@@ -452,9 +442,9 @@ ul#page-header-stats li.loading {
 }
 
 #pcgl-footer {
```
davidlougheed, line 453: not much point in using cat primary since this is pcgl-only anyway

src/js/components/SiteHeader.tsx
```diff
@@ -167,25 +165,30 @@ const SiteHeader = ({ menuItems }: SiteHeaderProps) => {
         </Flex>
 
         <Space size={isSmallScreen ? 4 : 'small'}>
```
davidlougheed, line 167: q: why did this order change?

src/js/components/Provenance/DatasetProvenance.tsx
```diff
@@ -310,7 +302,9 @@ export const DatasetProvenanceContent = ({ dataset }: { dataset: Dataset }) => {
       {dataset.long_description ? (
         <LongDescriptionBlock {...dataset.long_description} />
       ) : (
-        <Text italic>{t(dataset.description)}</Text>
+        <Paragraph ellipsis={{ rows: 5, expandable: true, symbol: 'more' }} style={{ fontStyle: 'italic', margin: 0 }}>
```
davidlougheed: use tailwindish class for italic/margin

src/js/components/Provenance/DatasetProvenance.tsx
```diff
@@ -324,7 +318,9 @@ export const DatasetProvenanceContent = ({ dataset }: { dataset: Dataset }) => {
         <Descriptions style={{ paddingTop: '20px' }}>
           {dataset.privacy && (
             <Item span={12} label={<DescLabel title={t('Privacy')} />}>
-              {t(dataset.privacy)}
+              <Paragraph ellipsis={{ rows: 2, expandable: true, symbol: 'more' }} style={{ margin: 0 }}>
```
davidlougheed: class `m-0` or `mb-0`, whichever properly eliminates margins according to need

src/js/components/Provenance/Catalogue/CatalogueToolbar.tsx
```diff
@@ -0,0 +1,117 @@
...
+    values.forEach((v) =>
+      pills.push({
+        key: `${facet}-${v}`,
+        facetLabel: FACET_LABELS[facet],
+        label: v,
+        onClose: () => dispatch(toggleFacetValue({ facet, value: v })),
+      })
+    );
+  });
+  if (q)
```
davidlougheed: i prefer wrapping this in curly brackets to avoid bugs

davidlougheed (changes requested):
code is looking better, although we will need a plan for these charts to be moved into bento public. I'm worried about the hard-coded charts/filters as tech debt, but since we need to deliver something for PCGL this is ok for now.

for now, mostly just small things in this code review.

also, i'm not convinced _all_ of the catalog variables need to be separate from antd, unless the idea is for these to become more broadly applied to all of Bento.

src/js/components/Provenance/Catalogue/CatalogueToolbar.tsx
SanjeevLakhwani: I prefer braceless as it feels similar to the implicit pattern that we do adopt in our codebase.

src/public/locales/en/default_translation_en.json
SanjeevLakhwani: Why?

src/js/components/Provenance/Catalogue/CatalogueToolbar.tsx
davidlougheed: It's a known anti pattern, albeit minor here: https://slate.com/technology/2014/02/apple-security-bug-a-critical-flaw-was-extraordinarily-simple.html

davidlougheed: https://dwheeler.com/essays/apple-goto-fail.html see 3.3

SanjeevLakhwani: fair

SanjeevLakhwani: if we are following this than maybe we can add this to linter also

src/js/components/SiteHeader.tsx
SanjeevLakhwani: as portal link is not permanent so pcgl on/off the sequence changes

PR #388, opened 2026-06-22
[WIP] add tracks view with igv.js

gsfk:
_... in progress_

Adds a "Tracks" tab to the existing phenopacket tabs:
- uses Bento references where they exist, falls back to igv
- similarly, prefers Bento feature search if present
- ignores case when searching for references to avoid existing reference headaches
- currently uses promises for initial track urls, which eliminates a few headaches but creates a few more

TODOs:
- [ ] test with crams  
- [x] test with multiple vcfs
- [ ] store igv position in redux
- [ ] replace "selected assembly" pulldown
- [ ] handle tracks with missing reference 
- [ ] lint


src/js/components/ClinPhen/TracksDisplay/TrackControlTable.tsx
```diff
@@ -0,0 +1,43 @@
...
+      title: t('general.view'),
+      key: 'view',
+      align: 'center' as const,
+      render: (_: unknown, track: ExperimentResultWithView) => (
+        <Switch checked={track.viewInIgv} onChange={() => toggleView(track)} />
+      ),
+    },
+  ];
+
+  return <Table bordered pagination={false} columns={trackTableColumns} rowKey="filename" dataSource={tracks} />;
```
davidlougheed, line 40: I think for consistency w/ rest of Bento Public, use the compact table display?

src/js/components/ClinPhen/TracksDisplay/TrackControlTable.tsx
```diff
@@ -0,0 +1,43 @@
...
+      title: t('experiment_result.description'),
+      dataIndex: 'description',
+    },
+    {
+      title: t('experiment_result.file_format'),
+      dataIndex: 'file_format',
+    },
+    {
+      title: t('experiment_result.genome_assembly_id'),
+      dataIndex: 'genome_assembly_id',
```
davidlougheed, line 28: use the genome assembly ID popover which is present in other parts of bento public (if ingested into the ref service)

src/js/components/ClinPhen/TracksDisplay/TracksView.tsx
```diff
@@ -0,0 +1,266 @@
...
+
+  const buildIgvTrack = useCallback(
+    (track: ExperimentResult): IgvTrack | null => {
+      if (!(track.url && track.file_format)) return null;
+
+      const igvFileDetails = caseInsensitiveIgvFileInfoLookup(track.file_format);
+      if (!igvFileDetails) return null;
+
+      if (!accessUrlsPromises) {
+        console.error('no access urls ');
```
davidlougheed, line 66: nitpick: rm space at the end of this string

src/js/hooks/igv.ts
```diff
@@ -0,0 +1,91 @@
...
+  return useAppSelector((state) => state.igv);
+};
+
+// get references in IGV format, preferring ones from bento when present
+export const useBentoOrIgvReferencesById = (requestedReferenceIds: string[]): IgvReferenceById => {
+  const { genomesStatus: bentoGenomeStatus, genomesByID: bentoReferenceGenomes } = useReference();
+  const { igvGenomesByID } = useIgvReference();
+
+  return useMemo(() => {
+    if (bentoGenomeStatus !== RequestStatus.Fulfilled) return {};
```
davidlougheed, line 22: could recover here if the request fails, relying on just IGV genomes

src/js/hooks/usePhenopacketTabs.tsx
```diff
@@ -24,6 +30,17 @@ export const usePhenopacketTabs = (phenopacket: Phenopacket | undefined) => {
     [navigate]
   );
 
+  const experimentResults = phenopacket ? phenopacketExperimentResults(phenopacket) : [];
+  const tracks = viewableTracks(experimentResults);
+  const requestedAssemblies = assemblyIdsForExperiments(tracks);
+  const referencesById = useBentoOrIgvReferencesById(requestedAssemblies);
```
davidlougheed, line 36: if this hook returns some kind of compound response status between igv + reference service, then we could fall back to using just igv.js genomes

davidlougheed (changes requested):
I get infinite loading when reference service errors:

Image

this is due to an error from the reference service in fact, but if there is an upstream error, the browser should ideally (not necessarily though) show some kind of error indication rather than an infinite-load. an in fact this could potentially be a recoverable error if we use the IGV.js provided references.

eventually this somehow splits into many IGV.js browsers, not sure how I triggered it though. I think it's because the cleanup callback isn't linked to the effect which creates the browsers, meaning the create can be called without a corresponding cleanup.

Image

gsfk:
- Now falls back to IGV correctly when reference service fails (previously only fell back when reference service was working but lacked a given assembly)
- Fixed infinite regress of IGV browser creation.
- IGV browser creation should now only fail in rare edge cases, eg:  the bento reference service works on first page load, but then fails later when trying to initialize IGV. 
  - in this case we just give a user notification of failure
  - Falling back to IGV reference in this case is difficult, since IGV returns a reference to the browser only on success. This can be resolved by a hard reload (which will force a fallback to IGV). Automatic fallbacks here are possible but inelegant, and largely equivalent to page reload. 
 - Added a French translation for the Tracks tab name ("Pistes"), although a possible title in both language is just "IGV"

src/js/hooks/igv.ts
gsfk: yes, this was much more strict than I intended

src/js/components/ClinPhen/TracksDisplay/TracksView.tsx
```diff
@@ -0,0 +1,270 @@
...
+    t,
+    tracksWithView,
+  ]);
+
+  // -------------------------- end igv init --------------------------
+
+  return (
+    <>
+      {(tracks.length > 0 || availableAssemblies.length > 0) && (
+        <div style={{ display: 'flex', flexDirection: 'column', gap: 10 }}>
```
davidlougheed: minor: use the antd Flex component instead

src/js/components/ClinPhen/TracksDisplay/TracksView.tsx
```diff
@@ -0,0 +1,270 @@
...
+                  igvContainerRefs.current[assemblyId] = node;
+                }}
+              />
+            </div>
+          ))}
+        </div>
+      )}
+      {availableAssemblies.map((assemblyId) => (
+        <div key={assemblyId} style={{ marginTop: 10 }}>
+          {hasMultipleAssemblies && <div style={{ fontWeight: 600, marginBottom: 5, marginTop: 15 }}>{assemblyId}</div>}
```
davidlougheed: minor: assign a heading level that fits with the page context here, and give it a slightly more descriptive name like `{{assemblyId}} track controls`

src/js/features/drs/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+export const getDrsObjectOrPassThrough = async (uri: string | undefined): Promise<DrsRecord | null> => {
+  /**
+   * If a record is returned, the URI is a DRS URI that has been fetched.
+   * If null is returned, the URI is not recognized as a DRS URI.
+   */
+
+  if (!uri) return null;
+
+  // deduplicate
+  const isDrs = (() => {
```
davidlougheed: can this be factored out into a helper fn? can be in this file, and then imported in the hooks file

src/js/hooks/usePhenopacketTabs.tsx
```diff
@@ -7,10 +7,16 @@ import OntologiesView from '@/components/ClinPhen/PhenopacketDisplay/OntologiesV
...
+import TracksView from '@/components/ClinPhen/TracksDisplay/TracksView';
 
 import { TabKeys } from '@/types/PhenopacketView.types';
 import type { Phenopacket } from '@/types/clinPhen/phenopacket';
 import { useTranslationFn } from '@/hooks';
+import { useScopeDownloadData } from '@/hooks/censorship';
+import { useReference } from '@/features/reference/hooks';
+import { assemblyIdsForExperiments, phenopacketExperimentResults } from '@/utils/experiments';
+import { useBentoOrIgvReferencesById, useIgvReference, viewableTracks } from './igv';
+import { RequestStatus } from '@/types/requests';
```
davidlougheed: nit: group with TabKeys const enum import

src/js/hooks/usePhenopacketTabs.tsx
```diff
@@ -24,6 +30,19 @@ export const usePhenopacketTabs = (phenopacket: Phenopacket | undefined) => {
...
+  const referencesById = useBentoOrIgvReferencesById(requestedAssemblies);
+  const { hasAttempted: attemptedCanDownload, hasPermission: canDownload } = useScopeDownloadData();
+  const { genomesStatus } = useReference();
+  const { igvGenomesStatus } = useIgvReference();
+
+  const tabsReady =
+    attemptedCanDownload &&
+    (tracks.length === 0 || genomesStatus === RequestStatus.Fulfilled || igvGenomesStatus === RequestStatus.Fulfilled);
+
+  // TODO: Add Experiments
```
davidlougheed: q: what does this mean?

src/js/components/ClinPhen/TracksDisplay/TracksView.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+      const fileUrls = accessUrlsPromises[track.url];
+      if (!fileUrls) {
+        console.error(`no access url for file`);
+        return null;
+      }
+
+      const { trackType, fileFormat } = igvFileDetails;
+
+      const t = {
```
davidlougheed: nitpick: arguably redundant variable

src/js/utils/igv.ts
```diff
@@ -0,0 +1,85 @@
...
+
+export const hasIndex = (track: ExperimentResult) => {
+  // index currently required but should probably be optional
+  return (track.indices ?? '').length > 0;
+};
+
+//  ---------------------------------------------
+// async equivalents of igv hooks
+
+const getDeferredDrsAccessMethod = (url: string | undefined): Promise<string | null> => {
```
davidlougheed: simple way to handle external files as well would be to rename this to be more generic, and then just return the original URL if the URL scheme is not `drs://`

src/js/utils/igv.ts
```diff
@@ -0,0 +1,85 @@
...
+  return track.indices.find((i) => acceptedIndicesThisType?.includes(i.format))?.url;
+};
+
+export const getIgvFileAndIndexAccessUrls = (tracks: ExperimentResult[]): IgvAccessUrlPromisesById => {
+  const drsUrls: Record<string, { fileAccessUrl: Promise<string | null>; indexAccessUrl?: Promise<string | null> }> =
+    {};
+
+  tracks.forEach((t) => {
+    if (!t.url) return;
+    const fileUrl = getDeferredDrsAccessMethod(t.url);
```
davidlougheed: the null case should be handled somewhere in here, so that fileUrl doesn't get passed as null to igv.js (which will produce a somewhat opaque error)

davidlougheed (changes requested):

src/js/hooks/usePhenopacketTabs.tsx
gsfk: if you mean `// TODO: Add Experiments` this was an existing comment in the code. Looks like it was removed after I branched off.

src/js/components/ClinPhen/TracksDisplay/TracksView.tsx
gsfk: `type` is "annotation", "wig", "alignment", "variant"  etc

`format` is "bam", "cram", "vcf" and so on. Easy to convert in most cases but I don't do it here, in part because there are extra steps needed to convert from bento file format which is all caps (eg "VCF"). 

One possible option is to eliminate the `type` field entirely, since igv can infer it from `format`.

davidlougheed: i meant `t` itself, since it's constructed and then immediately returned; other parts are fine to me

gsfk: ah, probably put there to avoid a TS freakout, I'll try without the declaration and see what happens.

src/js/components/ClinPhen/TracksDisplay/TracksView.tsx
gsfk: yeah, I only added labels where there was more than one assembly, I can add a generic title if it's not clear enough.

src/js/utils/igv.ts
gsfk: Are external files really in scope for a first round of IGV? If so there will have to be other changes (eg currently there are bento tokens on all calls)

davidlougheed: oh right, i guess igv will inject auth tokens no matter what...

gsfk: we can do tokens per-track, but may be hard/impossible to renew on-the-fly the way we currently do... so performance for native bento files may be worse.

gsfk: hmmm... perhaps not as bad as I'd thought, although I still suggest deferring this to later.

gsfk:
Now doesn't try to make tracks with null urls. 

I also made the behaviour consistent for external urls between the hook and non-hook version of the drs code. (Or, we can remove this from both if it's something we don't want).

PR #390, opened 2026-07-06
fix(catalogue): style fixes for footer, insights, and scope header

SanjeevLakhwani:
## Summary
- Darken PCGL footer background via `color-mix()` against `--ant-color-primary` instead of a hardcoded hex
- Use `text-secondary` for the catalogue insights header icon and title
- Remove padding from the scope header

### DIFF
--- a/src/js/components/Provenance/Catalogue/Charts/DonutChart.tsx
+++ b/src/js/components/Provenance/Catalogue/Charts/DonutChart.tsx
@@ -10,6 +10,23 @@ const SIZE = 116,
   R = 46,
   CIRC = 2 * Math.PI * R;
 
+const wrapCenterLabel = (label: string, maxLineLen = 10): string[] => {
+  const words = label.split(' ');
+  const lines: string[] = [];
+  let current = '';
+  for (const word of words) {
+    const candidate = current ? `${current} ${word}` : word;
+    if (candidate.length > maxLineLen && current) {
+      lines.push(current);
+      current = word;
+    } else {
+      current = candidate;
+    }
+  }
+  if (current) lines.push(current);
+  return lines;
+};
+
 interface DonutChartProps {
   title: string;
   data: { name: string; value: number }[];
@@ -47,6 +64,7 @@ const DonutChart = ({
   );
 
   const c = SIZE / 2;
+  const centerLabelLines = wrapCenterLabel(centerLabel);
 
   return (
     <Card size="small" className="chart-card">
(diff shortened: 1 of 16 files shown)


src/styles.css
```diff
@@ -570,7 +570,7 @@ body {
 .donut-wrap { display: flex; align-items: center; gap: 12px; }
 .donut { flex: none; }
 .donut-num { font-size: 18px; font-weight: 700; fill: rgba(0,0,0,0.88); text-anchor: middle; }
-.donut-lbl { font-size: 8px; fill: var(--text-muted); text-anchor: middle; text-transform: uppercase; letter-spacing: 0.05em; }
+.donut-lbl { font-size: 7px; fill: var(--text-muted); text-anchor: middle; text-transform: uppercase; letter-spacing: 0.04em; }
```
davidlougheed: would prefer staying at 8px - 7px is quite hard to read, especially with a lighter font weight.

src/js/components/Provenance/Dataset.tsx
```diff
@@ -110,7 +110,7 @@ const Dataset = ({
             ellipsis={{
               rows: 3,
               expandable: true,
-              symbol: <span className="catalogue-card__expand-symbol">more</span>,
+              symbol: <span className="catalogue-card__expand-symbol">{t('more', { count: 1 })}</span>,
```
davidlougheed: why a count for 'more'?

src/styles.css
```diff
@@ -124,6 +124,7 @@ body {
 .z-0 { z-index: 0; }
 .self-start { align-self: flex-start; }
 .text-cat-primary { color: var(--ant-color-primary, #054A74); }
```
davidlougheed: not a tailwind compatible class, same with the one below

src/js/components/Provenance/Catalogue/Charts/DonutChart.tsx
```diff
@@ -10,6 +10,23 @@ const SIZE = 116,
   R = 46,
   CIRC = 2 * Math.PI * R;
 
+const wrapCenterLabel = (label: string, maxLineLen = 10): string[] => {
```
davidlougheed: this seems brittle / a lot of code for text wrapping. I would prefer just using https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Element/foreignObject for text wrapping since we're in a web context rather than a print context.

davidlougheed (changes requested):

davidlougheed (approved):
lgtm, thanks for the fixes.

PR #392, opened 2026-07-09
Consolidate faceted search UI (catalogue + search sidebar)

SanjeevLakhwani:
- Made the URL the source of truth for catalogue filters (`q/sort/view/facet sets`) - Redux only hydrates from it now, components navigate instead of dispatching filter state directly. Mirrors how Search already works.
- Shared URL query-param plumbing (`utils/queryParams.ts`) between catalogue and search; catalogue facet params switched from comma-joined to repeated-key encoding to match search and fix a latent comma-splitting bug.
- Made the facet rail mobile-friendly : slide-over drawer below lg, toolbar collapses sort to an icon dropdown and hides the grid/list switch to keep search/sort/filters on one row.
- Extracted the URL-sync/action logic into generic hooks (useUrlFacetSync, useUrlFacetActions) and the rail's header+collapsible-sections chrome into a generic FacetSider component, then wired both CatalogueRail and SiteSider through them.
- Deduped a hardcoded facet-label translation map that duplicated key construction already done elsewhere.

Catalogue and search are two independent filter UIs (different Redux slices, different backing data, catalogue is static project/dataset metadata, search is live katsu discovery). This PR doesn't fully merge them, that's not viable yet since search's filters are dynamic/typed (enum/range/date) against katsu fields, while catalogue's are a fixed small set of string facets. As the logic would change significantly when we introduce Dataset's data based filtering, this would make it easier for us to combine both of them.

### DIFF
--- a/src/js/components/Util/FacetSider.tsx
+++ b/src/js/components/Util/FacetSider.tsx
@@ -1,0 +1,78 @@
+import type { ReactNode } from 'react';
+import clsx from 'clsx';
+import { CaretDownOutlined, CaretRightOutlined, CloseOutlined } from '@ant-design/icons';
+
+export interface FacetSiderSection {
+  id: string;
+  label: string;
+  collapsed: boolean;
+  onToggleCollapse: () => void;
+  /** Body content for the section, e.g. chips, a range input, a date picker — the shell doesn't care. */
+  render: () => ReactNode;
+}
+
+export interface FacetSiderProps {
+  title: string;
+  /** Shown in the header in place of the close button when not in overlay mode, e.g. a result count. */
+  countLabel?: ReactNode;
+  closeLabel: string;
+  /** Below some breakpoint, the sider renders as a fixed slide-over drawer instead of an inline column. */
+  overlay: boolean;
+  /** Ignored when `overlay` is false (the sider is always visible inline). */
+  open: boolean;
+  onClose: () => void;
+  sections: FacetSiderSection[];
+  /** BEM-style class prefix, e.g. "catalogue-rail" — drives `{prefix}`, `{prefix}-backdrop`,
+   *  `{prefix}--overlay`, `{prefix}--open`, and `{prefix}__*` header classes. */
+  classPrefix: string;
+}
+
+/**
+ * Generic facet-sidebar shell: header (title + count-or-close) plus a list of collapsible
+ * sections, with an optional overlay/slide-over presentation for narrow viewports. Each section
+ * owns its own collapse state and renders its own body, so the same shell works for the
+ * catalogue's chip facets and (eventually) the Search feature's enum/range/date filters.
+ */
+const FacetSider = ({
+  title,
+  countLabel,
+  closeLabel,
+  overlay,
+  open,
+  onClose,
+  sections,
+  classPrefix,
+}: FacetSiderProps) => (
+  <>
+    {overlay && open && <div className={`${classPrefix}-backdrop`} onClick={onClose} aria-hidden />}
+    <div className={clsx(classPrefix, overlay && `${classPrefix}--overlay`, open && `${classPrefix}--open`)}>
+      <div className={`${classPrefix}__header`}>
+        <span className={`${classPrefix}__title`}>{title}</span>
+        {overlay ? (
+          <button className={`${classPrefix}__close`} onClick={onClose} aria-label={closeLabel}>
+            <CloseOutlined />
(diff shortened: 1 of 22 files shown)


src/styles.css
```diff
@@ -589,14 +589,26 @@ body {
...
+.catalogue-sort-select { width: 180px; flex-shrink: 1; min-width: 70px; }
+.catalogue-toolbar-fixed { flex-shrink: 0; }
 .catalogue-count-highlight { color: rgba(0, 0, 0, 0.88); font-weight: 600; }
 .catalogue-filter-tag { border-radius: 13px !important; margin: 0 !important; }
 .catalogue-filter-tag__facet-label { color: var(--text-muted); font-size: 11px; margin-right: 3px; }
 .catalogue-clear-tag { border-radius: 13px !important; margin: 0 !important; cursor: pointer; border-style: dashed; }
 /* END catalogue toolbar */
 
+/* Bigger active-filter-tag variant used on the overview page (between the About section and count cards) */
+.overview-filter-tag {
```
davidlougheed: any way to avoid `!important` here to not break CSS inheritance? e.g., using the antd class for tag to increase specificity

src/styles.css
```diff
@@ -589,14 +589,26 @@ body {
...
+  font-size: 14px !important;
+  padding: 4px 10px !important;
+  height: auto !important;
+  line-height: 1.4 !important;
+}
+.overview-filter-tag .catalogue-filter-tag__facet-label {
+  font-size: 12.5px;
```
davidlougheed, line 599: going forward - it's preferable to use `rem`/`em` than fractional pixels in my opinion.

src/js/components/Util/FacetSider.tsx
```diff
@@ -0,0 +1,93 @@
...
+  overlay: boolean;
+  /** Ignored when `overlay` is false (the sider is always visible inline). */
+  open: boolean;
+  onClose?: () => void;
+  /** Renders as collapsible facet sections. Omit and use `children` for a non-sectioned body (e.g. a form). */
+  sections?: FacetSiderSection[];
+  children?: ReactNode;
+  /** BEM-style class prefix, e.g. "catalogue-rail" — drives `{prefix}`, `{prefix}-backdrop`,
+   *  `{prefix}--overlay`, `{prefix}--open`, and `{prefix}__*` header classes. */
+  classPrefix: string;
```
davidlougheed: I'd prefer CSS inheritance to programmatic class prefixing.

src/styles.css
SanjeevLakhwani: yeah but for rem/em we get a weird multi decimal number and em has the issue of inconsistent usage if changed now as it  can lead to weird behavior as it is relative to parent.

src/styles.css
```diff
@@ -592,14 +592,26 @@ body {
...
+.catalogue-sort-select { width: 180px; flex-shrink: 1; min-width: 70px; }
+.catalogue-toolbar-fixed { flex-shrink: 0; }
 .catalogue-count-highlight { color: rgba(0, 0, 0, 0.88); font-weight: 600; }
 .catalogue-filter-tag { border-radius: 13px !important; margin: 0 !important; }
 .catalogue-filter-tag__facet-label { color: var(--text-muted); font-size: 11px; margin-right: 3px; }
 .catalogue-clear-tag { border-radius: 13px !important; margin: 0 !important; cursor: pointer; border-style: dashed; }
 /* END catalogue toolbar */
 
+/* Bigger active-filter-tag variant used on the overview page (between the About section and count cards) */
+.ant-tag.overview-filter-tag {
```
davidlougheed, line 592: the border radius looks a bit off with the filter tags in the data overview - not enough radius to make it fully pill-shaped.

src/styles.css
```diff
@@ -592,14 +592,26 @@ body {
...
+  font-size: 14px;
+  padding: 4px 10px;
+  height: auto;
+  line-height: 1.4;
+}
+.overview-filter-tag .catalogue-filter-tag__facet-label {
+  font-size: 12.5px;
```
davidlougheed, line 599: looks odd to me with the 14px tag font size - already 'diminished' visually by being lighter, do we need it smaller as well?

SanjeevLakhwani: I increased cause we would need a bit off when smaller would require a bit of font size matching across the page. I shall make a ticket for it but I would suggest leaving it 14 for now to match other page elements.

PR #396, opened 2026-07-13
feat: add column picker for discovery match CSV export

SanjeevLakhwani:
Requires: https://github.com/bento-platform/katsu/pull/714
## Summary
- Add column picker for CSV export of search results, backed by new `discovery_matches_export_fields` endpoint
- New `ExportFieldsModal` lets users select a subset of columns before exporting; defaults to all fields (unchanged prior behavior)
- `useDownloadAllMatchesCSV` now accepts optional `fields` and sends `_fields` param only when a subset is chosen

### DIFF
--- a/src/js/components/Search/ExportFieldsModal.tsx
+++ b/src/js/components/Search/ExportFieldsModal.tsx
@@ -3,17 +3,19 @@ import { Checkbox, Divider, Flex, Modal, Space, Spin } from 'antd';
 
 import { useTranslationFn } from '@/hooks';
 import { useDiscoveryMatchExportFields } from '@/hooks/useDiscoveryMatchExportFields';
-import type { ResultsDataEntity } from '@/types/entities';
+import type { ExportFormat, ResultsDataEntity } from '@/types/entities';
 
 const ExportFieldsModal = ({
   open,
   entity,
+  format,
   exporting,
   onCancel,
   onExport,
 }: {
   open: boolean;
   entity: ResultsDataEntity;
+  format: ExportFormat;
   exporting: boolean;
   onCancel: () => void;
   onExport: (fields: string[] | undefined) => void;
@@ -51,8 +53,8 @@ const ExportFieldsModal = ({
     <Modal
       open={open}
       onCancel={onCancel}
-      title={t('search.export_csv')}
-      okText={t('search.export_csv')}
+      title={`${t('search.export_data')} (${t(`search.${format}`)})`}
+      okText={t('search.export')}
       onOk={() => onExport(allSelected ? undefined : [...selectedKeys])}
       okButtonProps={{ disabled: fetching || selectedKeys.size === 0, loading: exporting }}
     >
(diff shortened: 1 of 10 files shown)


src/js/hooks/useDiscoveryMatchExportFields.ts
```diff
@@ -0,0 +1,42 @@
+import { useCallback, useState } from 'react';
+import axios from 'axios';
+import { useAppSelector } from '@/hooks';
+import { useSelectedScope } from '@/features/metadata/hooks';
+import { scopedAuthorizedRequestConfigFromParts } from '@/utils/requests';
+import { katsuDiscoveryMatchesExportFieldsUrl } from '@/constants/configConstants';
+import type { ResultsDataEntity, ExportField } from '@/types/entities';
+
+// Fields available per entity rarely change within a session, so cache per-entity once fetched.
+const fieldsCache = new Map<ResultsDataEntity, ExportField[]>();
```
davidlougheed: here, can we just use redux for consistency since this is now caching between fetches? otherwise we now have a third pattern of state management in the same codebase (local, local + cache, redux cache)

src/js/components/Search/ExportFieldsModal.tsx
```diff
@@ -0,0 +1,91 @@
...
+            </Checkbox>
+            <span className="text-secondary">
+              {selectedKeys.size} / {fields.length}
+            </span>
+          </Flex>
+          <Divider className="my-3" />
+          <Space direction="vertical">
+            {fields.map(({ key, label }) => (
+              <Checkbox key={key} checked={selectedKeys.has(key)} onChange={(e) => toggleKey(key, e.target.checked)}>
+                {label}
```
davidlougheed: I think I would prefer rendering the field ID directly in monospace font rather than trying to transform it into text via split/title-casing. alternatively, use the field IDs to lookup into a translation dictionary (there are already objects in the localization for biosample I believe, for example) to translate the field name nicely.

src/js/components/Search/ExportFieldsModal.tsx
```diff
@@ -0,0 +1,91 @@
...
+              {t('search.select_all_fields')}
+            </Checkbox>
+            <span className="text-secondary">
+              {selectedKeys.size} / {fields.length}
+            </span>
+          </Flex>
+          <Divider className="my-3" />
+          <Space direction="vertical">
+            {fields.map(({ key, label }) => (
+              <Checkbox key={key} checked={selectedKeys.has(key)} onChange={(e) => toggleKey(key, e.target.checked)}>
```
davidlougheed: use a CheckboxGroup perhaps? would be slightly cleaner

src/js/hooks/useDiscoveryMatchExportFields.ts
davidlougheed: that would also allow some error handling for this since we already have logic for it

src/js/components/Search/ExportFieldsModal.tsx
SanjeevLakhwani: The fields are derived from Katsu, as discussed.

